PR #194, opened 2014-12-30
User id being set for Google API

robstrong:
This fixes an issue with #187 where the user id was not being set because the request for the user details did not specifically ask for it.


sportarc:
Thanks for the fix. I just realized this issue and was looking for a fix.

ramsey:
Thanks for the fix!

PR #196, opened 2015-01-06
Updated README.md to include Google Nest provider

JC5:
I've written a simple yet effective Google Nest provider. It includes an object with all supported API fields and devices.


ramsey:
Thanks for adding it to the list!

PR #198, opened 2015-01-08
Fix UserTest.php namespace

re-24:


ramsey:
Thanks for fixing this!

PR #199, opened 2015-01-08
Align all array items

jbrooksuk:
A handful of arrays had been aligned, this just makes them all the same.

### DIFF
--- a/src/Entity/User.php
+++ b/src/Entity/User.php
@@ -53,18 +53,18 @@ class User
     public function getArrayCopy()
     {
         return [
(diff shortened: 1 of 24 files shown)


jbrooksuk:
Ping.

ramsey:
The build is failing. Additionally, I do not agree with vertical alignment for reasons that are described here: http://en.wikipedia.org/wiki/Programming_style#Vertical_alignment

Thanks for your eagerness to contribute, but I'm declining this pull request. Sorry.

PR #201, opened 2015-01-09
Add FreeAgent Third Party Provider to the documentation

### LINKED ISSUE #200: Authorization Bearer Header
Hi there!

I'm trying to integrate [Freeagent](https://dev.freeagent.com/docs) as a provider to your OAuth2 library however they send the token on each request in the headers, so my idea was override your method [fetchProviderData](https://github.com/thephpleague/oauth2-client/blob/374c3d41db38d1d2d4071c58fa5515f58ea4566e/src/Provider/AbstractProvider.php#L317) to be able to receive the token in that method and add it to the `header` as below:

```
protected function fetchProviderData($url, \League\OAuth2\Client\Token\AccessToken $token)
    {
        try {
            $client = $this->getHttpClient();
            $client->setBaseUrl($url);

            $client->setDefaultOption('headers', ['Authorization' => 'Bearer ' . $token]);

            $request = $client->get()->send();
```

Do you think it will be a good idea send a PR to change your [fetchProviderData](https://github.com/thephpleague/oauth2-client/blob/374c3d41db38d1d2d4071c58fa5515f58ea4566e/src/Provider/AbstractProvider.php#L317) method? for something like this:

```
protected function fetchProviderData($url, \League\OAuth2\Client\Token\AccessToken $token = null)
``` 

So we will be able to pass if we need it in our implementation the token.

Let me know your thoughts!

Cheers

zot24:
Close #200

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,7 @@ so please help them out with a pull request if you notice this.
 - [Vkontakte](https://packagist.org/packages/j4k/oauth2-vkontakte)
 - [Naver](https://packagist.org/packages/deminoth/oauth2-naver)
 - [Google Nest](https://github.com/JC5/nest-oauth2-provider)
+- [FreeAgent](https://github.com/CloudManaged/oauth2-freeagent)
 
 ### Implementing your own provider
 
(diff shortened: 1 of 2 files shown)


ramsey:
Thanks for your contribution!

PR #207, opened 2015-01-20
Add support for specifying the access_type parameter for Google OAuth

AnduAanei:
In order to receive a `refreshToken` from Google Auth, `access_type` must be set to `offline` in the authorization url. The Google provider didn't allow setting this parameter.


ramsey:
Thanks for your contribution!

PR #217, opened 2015-02-02
Changes to additional providers

tpavlek:
Added Twitch to the additional providers, and alphabetized the list for more grokkability.


ramsey:
Thanks for your contribution!

Issue #193, opened 2014-12-30
PHP errors on /oauth2callback for Google sign-in


sylouuu:
Hi,

I'm using Codeigniter.

I'm sign-in from `/auth/google` and I added your code to the controller, I just updated the provider array:

```php
$provider = new League\OAuth2\Client\Provider\Google(array(
    'clientId'  =>  '...',
    'clientSecret'  =>  '...',
    'redirectUri'   =>  '..',
    'scopes' => array('email')
));

if (!isset($_GET['code'])) {

    // If we don't have an authorization code then get one
    $authUrl = $provider->getAuthorizationUrl();
    $_SESSION['oauth2state'] = $provider->state;
    header('Location: '.$authUrl);
    exit;

// Check given state against previously stored one to mitigate CSRF attack
} elseif (empty($_GET['state']) || ($_GET['state'] !== $_SESSION['oauth2state'])) {

    unset($_SESSION['oauth2state']);
    exit('Invalid state');

} else {

    // Try to get an access token (using the authorization code grant)
    $token = $provider->getAccessToken('authorization_code', [
        'code' => $_GET['code']
    ]);

    // If you are using Eventbrite you will need to add the grant_type parameter (see below)
    $token = $provider->getAccessToken('authorization_code', [
        'code' => $_GET['code'],
        'grant_type' => 'authorization_code'
    ]);

    // Optional: Now you have a token you can look up a users profile data
    try {

        // We got an access token, let's now get the user's details
        $userDetails = $provider->getUserDetails($token);

        // Use these details to create a new profile
        printf('Hello %s!', $userDetails->firstName);

    } catch (Exception $e) {

        // Failed to get user details
        exit('Oh dear...');
    }

    // Use this to interact with an API on the users behalf
    echo $token->accessToken;

    // Use this to get a new access token if the old one expires
    echo $token->refreshToken;

    // Number of seconds until the access token will expire, and need refreshing
    echo $token->expires;
}
```

I have the confirm box, I accept. I'm redirected to `/oauth2callback` as expected. I've added the same code (just for testing) in that controller. `$_GET['state']` and `$_GET['code']` are filled.

Error is:

```
A PHP Error was encountered

Severity: 4096

Message: Argument 1 passed to League\OAuth2\Client\Provider\AbstractProvider::setResultUid() must be of the type array, null given, called in /path/to/vendor/league/oauth2-client/src/Provider/AbstractProvider.php on line 221 and defined

[...]

PR #218, opened 2015-02-05
Get reliable body response on error.

dolfelt:
Certain responses (pretty JSON) cause issues when you try to explode the raw response. This affects the Google provider in my experience. Using the `getBody` method should be more reliable.

Might be related to #193.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -196,8 +196,7 @@ abstract class AbstractProvider implements ProviderInterface
             }
         } catch (BadResponseException $e) {
             // @codeCoverageIgnoreStart
-            $raw_response = explode("\n", $e->getResponse());
-            $response = end($raw_response);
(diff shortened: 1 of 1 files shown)


dolfelt:
Any thoughts on this @ramsey or @alexbilbie? This is causing breaking issues for me when using the Google+ provider.

ramsey:
Sorry, @dolfelt. I was traveling when you submitted this and haven't been able to look yet. This fixing the breaking issues you were seeing?

dolfelt:
@ramsey Yep. The problem is that the error response coming back from Google is formatted JSON. The JSON itself has line breaks in it.

When using this:
```php
$raw_response = explode("\n", $e->getResponse());
$response = end($raw_response);
```
It causes the `$response` object to come back as `}`, rather than the whole JSON object.

ramsey:
Thanks, @dolfelt!

Issue #216, opened 2015-02-01
Is there any way to disable approval_prompt


mrgodhani:
Hi

I am trying to implement Dropbox oauth2 using this package. But unfortunately dropbox doesn't accept approval_prompt as one of their fields. So is their any way to disable that without manually digging into the code?

ramsey:
This was resolved in #220, so closing this issue.

Issue #40, opened 2013-11-24
Problems adding DropBox as a provider


jnbdz:
I tried to add DropBox as a provider.

Here is my code: https://gist.github.com/jnbdz/7624654

I was actually able to solve the problem quickly, but I think this is something you guys should think about for your next version.

To make DropBox work I had to comment out line 70:

    //'approval_prompt' => 'force' // - google force-recheck

The problem then is if I want to use Google it won't work anymore...

ramsey:
I believe PR #220 has fixed this. If it's still not fixed, please reopen this issue.

PR #220, opened 2015-02-12
Allow approval_prompt to be set through options

jasonvarga:
Dropbox OAuth authorization URL fails if `approval_prompt` is in the query string. See screenshot:

image

This PR allows a provider to override it. This addresses #216 and is adapted from #40.
This allows you to do this in your provider:

``` php
public function getAuthorizationUrl($options = array())
{
  return parent::getAuthorizationUrl(array_merge([
    'approval_prompt' => []
  ], $options));
}
```

Setting the value to `[]` will cause it to be removed when `httpBuildQuery` is used in the parent.

I've created a Dropbox provider that I'll be moving into its own package. It was originally going to be part of this PR, but has since been removed, so you can see it in the commit history for this PR if you want.

If this gets the OK, I'll open a separate PR to link to my provider.


ramsey:
Looks good. Thanks for the pull request!

jasonvarga:
Does this need to be tagged before we can use this update?

ramsey:
It does need to be tagged. I was considering some other issues before tagging it, though.

jasonvarga:
Sure, no problem. Thanks for merging.

ramsey:
I've just cut the 0.8.1 tag, which includes this fix. Thanks!

PR #221, opened 2015-02-12
Add link to Dropbox provider

jasonvarga:


ramsey:
Thanks! Might want to use `~0.8.1` in your composer.json, rather than `~0.8`, just in case some project somewhere hard locks oauth2-client to `0.8.0` and tries to use the Dropbox provider.

jasonvarga:
Good call, done. Thanks!

philsturgeon:
The larger concern with `~0.8` is that 0.9 can have BC breaks according to SemVer, and 1.0 definitely will. While you require any 0.x version always be specific with `0.8.*` or `0.8.1`. 

That is what @ramsey suggested, but for a different reason. :)

jasonvarga:
Makes sense. Thanks Phil.

PR #223, opened 2015-02-20
Remove Guzzle 3.x dependency (without BC)

jeromegamez:
Greetings!

The changes in this PR enable developers to override the Guzzle 3.x dependency with any library supported by the [Ivory HTTP adapter](https://github.com/egeloen/ivory-http-adapter) (which, as a bonus point, implements PSR-7).

I lied in the PR title :) - in order not to introduce breaking changes for existing users, the default HTTP adapter is still using the Guzzle ~3.7 dependency. This is to address the possible use cases where projects rely on `thephpleague/oauth2-client` to require `"guzzle/guzzle": "~3.7"` and use it for their own needs.

However, if a project is using another HTTP client, it can be configured like this:

```php
use Ivory\HttpAdapter\CurlHttpAdapter;
use Ivory\HttpAdapter\GuzzleHttpHttpAdapter;

$httpClient = new CurlHttpAdapter(); // Native curl
// or
$httpClient = new GuzzleHttpHttpAdapter(); // Guzzle 5
// or …

$provider = new League\OAuth2\Client\Provider\<ProviderName>([
    'clientId'      => 'XXXXXXXX',
    'clientSecret'  => 'XXXXXXXX',
    'redirectUri'   => 'https://your-registered-redirect-uri/',
    'scopes'        => ['email', '...', '...'],
], $httpClient);
```

You can find a list of supported adapters and a full usage documentation in the [Ivory HTTP Adapter README](https://github.com/egeloen/ivory-http-adapter/blob/master/README.md).

I hope this PR receives your approval, and your feedback would be highly appreciated.

Cheers
:octocat: Jérôme

PS: If breaking changes are okay (as the oauth2-client library is at a 0.x release), we could also replace the default HTTP adapter to the Curl Adapter and completely remove the Guzzle 3.x dependency.


ramsey:
Version 1.0 of oauth2-client has been upgraded to use Guzzle 5. I would prefer to make these kinds of changes in the 1.0 branch rather than in the 0.x series.

Do you mind taking a look at the 1.0 branch and seeing if this work is more appropriately applied there?

jeromegamez:
Thank you for your quick feedback @ramsey, I just created a new PR, targeted at the 1.0 branch, at #224.

PR #224, opened 2015-02-20
Remove Guzzle as a dependency and allow custom HTTP clients

jeromegamez:
_This is a follow-up to #223._

Greetings!

The changes in this PR remove Guzzle as a dependency and allows developer to use their already existing HTTP client as long as it is supported by the [Ivory HTTP adapter](https://github.com/egeloen/ivory-http-adapter) (which, as a bonus point, implements PSR-7).

In the default state, the `CurlHttpAdapter` is used (which only has `curl` as a dependency), but if a project is already using another HTTP client, it can be configured like this:

```php
use Ivory\HttpAdapter\GuzzleHttpHttpAdapter;

$httpClient = new GuzzleHttpHttpAdapter(); // Guzzle 5

$provider = new League\OAuth2\Client\Provider\<ProviderName>([
    'clientId'      => 'XXXXXXXX',
    'clientSecret'  => 'XXXXXXXX',
    'redirectUri'   => 'https://your-registered-redirect-uri/',
    'scopes'        => ['email', '...', '...'],
], $httpClient);
```

You can find a list of supported adapters and a full usage documentation in the [Ivory HTTP Adapter README](https://github.com/egeloen/ivory-http-adapter/blob/master/README.md).

I hope this PR receives your approval, and your feedback would be highly appreciated.

Cheers
:octocat: Jérôme


jeromegamez:
Oh, and in case you accept this PR, I will always create PRs at the 3rd party libraries currently using the 1.0 branch of the oauth2-client. As far as I can tell, that's currently only the case for the [oauth2-facebook](https://github.com/thephpleague/oauth2-facebook) provider.

src/Provider/AbstractProvider.php
```diff
@@ -323,17 +321,12 @@ protected function fetchProviderData($url)
...
-            }
-
-            $request = $client->get($url, $options);
-            $response = $request->getBody();
-        } catch (BadResponseException $e) {
+            $response = (string) $httpResponse->getBody();
+        } catch (HttpAdapterException $e) {
             // @codeCoverageIgnoreStart
-            $raw_response = explode("\n", $e->getResponse());
+            $raw_response = explode("\n", (string) $e->getResponse()->getBody());
```
ramsey: I believe this has changed in recent 0.x code. I need to merge changes from master into the 1.0 branch, and then we'll see if there are any conflicts.

jeromegamez: As soon as you have merged the 0.x into the 1.0 branch, I can rebase the PR and handle conflicts, if there are any. But this PR is very similar to #223 (which I targeted at 0.x) so I have a good feeling :)

ramsey:
I merged master into 1.0. It now looks like your PR has some conflicts, and the 1.0 branch is now failing the build (see https://travis-ci.org/thephpleague/oauth2-client/builds/52007665).

The build failures are really on me, so I'll handle those, but it will be a day or so before I can get to them.

ramsey:
I fixed the build. Please rebase and see what conflicts there are.

jeromegamez:
I rebased and resolved the one conflict that the rebase produced.

ramsey:
Thanks!

PR #225, opened 2015-02-23
Clean out custom provider code from AccessToken

shadowhand:
Mail.ru and Battle.net have been moved out to separate repos and already resolved this. Vkontakte should use `$uidKey`.


ramsey:
Thanks for your contribution!

PR #226, opened 2015-02-23
Add additional method for modifying token response

shadowhand:
Create `AbstractProvider::prepareAccessTokenResult` to allow providers to modify token responses. Particularly useful for mapping expirations and capturing any additional details that may come back with the token.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -215,9 +215,22 @@ abstract class AbstractProvider implements ProviderInterface
             // @codeCoverageIgnoreEnd
         }
 
-        $this->setResultUid($result);
+        $result = $this->prepareAccessTokenResult($result);
 
         return $grant->handleResponse($result);
+    }
+
+    /**
+     * Prepare the access token response for the grant. Custom mapping of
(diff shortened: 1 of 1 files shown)


shadowhand:
In some ways, it would be helpful if `AccessToken` simply took all additional fields passed in `$options` and stored them (perhaps a `$details = []` property?) for later usage.

ramsey:
Thanks for the pull request!

Having a `$details` property on the AccessToken sounds like a good idea. In the 1.0 branch, though, I want to move away from public properties, so it might be a good idea to provider some accessor method to the `$details` property, rather than making it public.

shadowhand:
@ramsey any idea when a new release will be tagged? This PR in particular is preventing our Square provider from being completely functional, because Square returns tokens with `expires_at` field, which we need to remap to `expires` to be recognized by the AccessToken class.

Thanks for the quick response!

ramsey:
I've just tagged the 0.9.0 release. Enjoy!

shadowhand:
Thanks Ben, that's great! Will update the Square provider asap.

PR #228, opened 2015-02-24
Allow provider to add some extra user info to user class

binaryoung:
Some sites not only provides this user info to oauth client,but provider can't add extra info to user class

### DIFF
--- a/src/Entity/User.php
+++ b/src/Entity/User.php
@@ -32,13 +32,13 @@ class User
 
     public function __set($property, $value)
     {
-        if (!property_exists($this, $property)) {
+        /*if (!property_exists($this, $property)) {
             throw new \OutOfRangeException(sprintf(
                 '%s does not contain a property by the name of "%s"',
                 __CLASS__,
                 $property
             ));
-        }
+        }*/
 
         $this->$property = $value;
 


ramsey:
Please take a look at our [contributing guidelines](https://github.com/thephpleague/oauth2-client/blob/master/CONTRIBUTING.md).

This pull request breaks the build. Additionally, please do not commit commented-out code to the repository.

If you could provide more information about the problem you are having, perhaps there is another work-around that does not involve removing this code. As it is right now, it looks like you're running into an issue with a provider that is trying to populate a property on the `User` that does not exist. Is this correct?

binaryoung:
Yes,i want to Implement my own provider.But when i try to populate a property on the User that does not exist,i will catch a error.Can you tell me some ways to solve this problem?Thanks.

ramsey:
Can you provide examples of the properties you are attempting to set on the User. Also, what OAuth 2 server are you attempting to authenticate with that is passing these properties to you?

binaryoung:
For example,baidu oauth2 server will provide marriage status,blood type,constellation,education,job info of user.And qq(Chinese largest company of social website and IM software)will provider user's level and fans number info.Also sina will provide friends Count,created time,favourites Count,followers Count info.So i think it's necessary to allow provider to populate a property on the User that does not exist.Any idea?

ramsey:
To be honest, I'm not even sure why this library provides a User object. The User is not part of the OAuth 2 specification.

@philsturgeon, @alexbilbie, can you provide some background on the need for the User class? I think we'll continue to run into these issues where each server has different User needs. Since User isn't a part of the OAuth 2.0 spec, my opinion is that it shouldn't be a part of this library. Thoughts?

philsturgeon:
The user is definitively outside of the oauth2 specification. This is one of the assumption being made waaay way back when it was just CI code, and all people wanted to do was make user logins to get their Facebook user details and make a profile. 

The idea was to abstract as much of the provider stuff as possible in doing that, so you could write one chunk of code and have all of your oauth2 logins sorted out, and swap that for user data. Things like uid, username, etc was so different in so many systems that this had to be there, just to make lives easier for people during this process, as that was the whole goal of the app. 

Since then it grew to support more grant types, and has obviously made some moves away from this use case and onwards to being a generic oauth2 client that just does what the spec does. It could be removed, but this package would drop a lot of its use if it did. 

If it could be split somehow then that would be really handy, but the providers would not know how to provide this functionality and we'd be stuck in a really weird position.

tpavlek:
I'm definitely in favour of keeping the `User` objects around. This app shouldn't provide API functionality, but if it's authenticating a user, I do not think it's unreasonable to have it inform who that user is, regardless of what is strictly in the spec.

I solved this problem in my Battle.net package by just adding a `BattleNetUser` class, and having the `->userDetails()` function in my provider return that: https://github.com/tpavlek/oauth2-bnet/blob/master/src/Provider/BattleNetUser.php.

The provider has to know about what data it's going to be pulling anyway, so I don't see any issues with providers just doing something similar.

ramsey:
@tpavlek I like how you've solved this in your package. IMO, the User is really a provider-specific construct. I don't mind keeping the base class around as a generic User object, but I'd prefer not to make too many assumptions about it.

@binaryoung, does @tpavlek's solution help you? Is that something you can do in your providers, without making upstream changes?

binaryoung:
Thanks guys,thanks for your great solutions.It really helped me ;)

Issue #106, opened 2014-06-09
Make error_description and error_uri accessible


aripringle:
The OAuth2 spec allows for error_description and error_uri parameters in the error response (http://tools.ietf.org/html/rfc6749#section-5.2). The current implementation of IDPException doesn't expose these at all.

These details could arguably be added to the exception message, but I could see a use case for keeping the message with just the "error" key so it can be easily parsed. Would it make sense to add getErrorDescription and getErrorUri methods? Or perhaps just a getResult method that provides access to the $result property?

ramsey:
Implemented in #229.

PR #229, opened 2015-02-26
Added getResponseBody method in IDPException

### LINKED ISSUE #185: Fetching server response from IDPException
I'm creating provider for Twitch and when authorization code is invalid (for example) it will return "Bad Request" as error. When getting message out of IDPException it will return the "Bad Request". However inside IDPException is protected property result that will contain additional information about the error. This additional data should be fetchable from IDPException.
```
'result' => [
    'error' => 'Bad Request',
    'status' => 400
    'message' => 'Invalid authorization code'
]
```
This message part is unfetchable right now. I would suggest creating new method inside IDPException to fetch actual server response.

vimishor:
### Problem:

Some providers are sending additional information when an error occurs, which can't be accessed because property `$result` is protected.

### Solution:

Data from this property can be accessed using `getResponseBody()` method.

Closes #185 and maybe #106 .

### DIFF
--- a/src/Exception/IDPException.php
+++ b/src/Exception/IDPException.php
@@ -25,6 +25,11 @@ class IDPException extends \Exception
         parent::__construct($message, $code);
     }
 
+    public function getResponseBody()
+    {
+        return $this->result;
+    }
+
     public function getType()
     {
         $result = 'Exception';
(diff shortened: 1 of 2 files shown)


ramsey:
Thanks for your contribution!

PR #232, opened 2015-03-01
Enhance ability to use client to make authenticated requests

shadowhand:
- add `ProviderInterface::getHeaders()`
- use `getHeaders()` for `fetchUserDetails()`

Some providers do not provide a method for using query strings to send tokens.
By adding hooks for headers, the provider can be used to sign API requests,
including POST requests.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -31,7 +31,9 @@ abstract class AbstractProvider implements ProviderInterface
 
     public $responseType = 'json';
 
-    public $headers = null;
+    public $headers = [];
+
+    public $authorizationHeader;
 
     /**
      * @var GuzzleClient
@@ -323,17 +325,19 @@ abstract class AbstractProvider implements ProviderInterface
     {
         $url = $this->urlUserDetails($token);
 
-        return $this->fetchProviderData($url);
+        $headers = $this->getHeaders($token);
+
+        return $this->fetchProviderData($url, $headers);
     }
 
-    protected function fetchProviderData($url)
+    protected function fetchProviderData($url, array $headers = [])
     {
         try {
             $client = $this->getHttpClient();
             $client->setBaseUrl($url);
 
-            if ($this->headers) {
-                $client->setDefaultOption('headers', $this->headers);
+            if ($headers) {
+                $client->setDefaultOption('headers', $headers);
             }
 
             $request = $client->get()->send();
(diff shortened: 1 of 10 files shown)


shadowhand:
I wasn't sure what to do about the Microsoft provider, the only reference I found for an `Authorization` header was in [this article about CRM](http://blogs.msdn.com/b/crm/archive/2013/12/12/use-oauth-to-authenticate-with-the-crm-service.aspx) but I am unsure if that applies to all of Microsofts APIs.

shadowhand:
@ramsey if possible, I would really appreciate if this could be looked at today, as this is a critical part of our ability to use the League OAuth clients. Same goes for thephpleague/oauth1-client#18. Thank you!

ramsey:
I am traveling. I will try my best to look at it today.

src/Provider/AbstractProvider.php
```diff
@@ -323,17 +325,19 @@ protected function fetchUserDetails(AccessToken $token)
...
         $url = $this->urlUserDetails($token);
 
-        return $this->fetchProviderData($url);
+        $headers = $this->getHeaders($token);
+
+        return $this->fetchProviderData($url, $headers);
     }
 
-    protected function fetchProviderData($url)
+    protected function fetchProviderData($url, array $headers)
```
ramsey: Should we considering initializing this to a default value in case any providers in the wild are calling the `fetchProviderData()` method? Perhaps something as simple as this:

    protected function fetchProviderData($url, array $headers = [])

src/Provider/AbstractProvider.php
```diff
@@ -31,7 +31,9 @@
 
     public $responseType = 'json';
 
-    public $headers = null;
+    public $headers = [];
+
+    public $authorizationHeader;
```
ramsey: Will existing third-party providers need to update themselves to support this addition?

src/Provider/AbstractProvider.php
shadowhand: That's a good idea. Will Change.

src/Provider/AbstractProvider.php
shadowhand: Yes.

src/Provider/AbstractProvider.php
shadowhand: @ramsey added the default.

src/Provider/AbstractProvider.php
ramsey: What is the impact if I merge this in now and tag a release? How badly will providers break if they are using oauth2-client `~0.9` (as they would be automatically upgraded to `0.10` if using versioning like this)?

shadowhand: For the default providers that ship with this package? No impact, everything should be complete transparent.

For a custom provider, everything should be smooth, unless `fetchProviderData` has been heavily modified and the code somehow became incompatible. I tried my best to make sure that existing code would not break with these changes applied... I don't think any of the abstract changes will break existing functionality.

ramsey: I see. If a custom provider continues to use the access token as we've been using it, then nothing is affected; they won't need to mess with `$authorizationHeader` unless they want to.

ramsey:
Thanks!

src/Provider/AbstractProvider.php
shadowhand: > continues to use the access token as we've been using it, then nothing is affected

Exactly. Thanks for merging!

PR #233, opened 2015-03-01
Stop using deprecated Travis option --dev

shadowhand:
As per recent Travis builds:

> You are using the deprecated option "dev". Dev packages are installed by default now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,8 @@ php:
 
 before_script:
   - travis_retry composer self-update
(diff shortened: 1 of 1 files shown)


ramsey:
The `--dev` option is actually a Composer option and not related to Travis. While `--dev` is enabled by default in Composer, I was unaware that it had been deprecated.

According to this commit, the Composer project added a deprecated warning for `--dev` on December 7:
https://github.com/composer/composer/commit/32218e97d14432d657e23faac42571546482f1ff

Thanks for pointing this out. Merging in your PR.

Issue #234, opened 2015-03-05
Please make it php5.3.* compatible


peterdd:
Composer installs only 0.3 version of this library, because all version after that require php5.4.0 or later.

But that version has problems with lower/uppercase name.

php cli 5.3.3
php web 5.3.28

My workaround is
  
    cp AuthorizationCode.php  Authorizationcode.php

But that is no solution for a new Flyspray release we plan.

ramsey:
One of the League requirements is that the packages be PHP 5.4+. Additionally, the class name is correctly `AuthorizationCode`. Why do you need to make the "C" lowercase?

tpavlek:
The bug he's referring to is one that was around before you took over @ramsey. There was an open bug where it didn't properly uppercase the C in authorization code for the default grant type.

@peterdd since 5.4+ is the minimum for the package now, your only solution is the one you presented, or you can fork the project and make changes to `AbstractProvider`. You want to change the getAccessToken method to include this line

```php
$className = str_replace(' ', '', ucwords(str_replace(['-', '_'], ' ', $grant)));
```

instead of the old, incorrect casing rules.

But those are your only two solutions unless you upgrade PHP.

rakeev:
Why not simply fix classname capitalization and tag as 0.3.1?

peterdd:
The current active devs at https://github.com/Flyspray/flyspray aren't experienced with that composer stuff. php5.2.* and php5.3.* is still wide spread on hostings and we want Flyspray be very easy to install and userfriendly.

This oauth2-client seems so far the only showstopper on php5.2.* or php5.3.* environments, so if you know a solution that we can until the release we plan around end this month we would be very glad.
Maybe some composer.json modification? Well, we just don't know to less about composer.

@rakeev: If that is possible and the composer can handle that automatically that would be great!

tpavlek:
I'm not sure if @ramsey wants to tag, but I went ahead and fixed this over here: https://github.com/tpavlek/oauth2-client.

@peterdd here is how you can fix this for your projects. First, go to the above link, I've already performed the capitalization fix and tagged a 0.3.1 release. In the upper right hand corner, click fork, and create a new fork of that project in your account.

You should now be on https://github.com/peterdd/oauth2-client or something similar.

You can now instruct any applications that use oauth2-client to use *your* fork instead of the official PHP League one. All you have to do is in the `composer.json` of the project requiring PHP 5.3 support add a section:

```json
"repositories": [
    {
[...]

PR #235, opened 2015-03-09
PHP5.3 compatible

peterdd:
I don't know if its currently enough justs to replace short array tags with array().

But it is currently a showstopper to have oauth2-client not php5.3 compatible.

If you don't like php5.3 compatibilty in your master branch, please add a php5.3 branch thats created and updated to composer on every release too.

The fix of the camelcase name bug in version 0.3 is not included. I think the better way is to update current version to support also php5.3 and let php5.3 users profit from current oauth2-client development.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: php
 
 php:
+  - 5.3
   - 5.4
   - 5.5
   - 5.6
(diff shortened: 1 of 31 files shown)


ramsey:
We're not going to back-port this library to work with PHP 5.3. We can't commit to support versions of PHP older than 5.4. Sorry.

As discussed in #234, you may fork the repository and maintain your own version that is compatible with PHP 5.3.

PR #236, opened 2015-03-11
[Fix] Invalid JSON in response will trigger fatal error

### LINKED ISSUE #230: Invalid server response crashes client.
Hi,

If the client connects to a server that gives an invalid response, the client crashes.

```
Catchable fatal error: Argument 1 passed to League\OAuth2\Client\Provider\AbstractProvider::prepareAccessTokenResult() must be of the type array, null given, called in .../oauth2/vendor/league/oauth2-client/src/Provider/AbstractProvider.php on line 218 and defined in .../oauth2/vendor/league/oauth2-client/src/Provider/AbstractProvider.php on line 230
```

If you consider this a bug I would be happy to create a pull request.

vimishor:
If the server responded with an invalid JSON for `getAccessToken()`,
$result variable will be of type string instead of expected array.
This will result in a fatal error was being raised in AbstractProvider
when calling `prepareAccessTokenResult` method, which expects an
argument of type array.

Proposed solution is to set the variable `$result` to an empty array if
json_decode was not successful.

Doesn't make sense to throw an expection at this point, because response
will not have an access_token and an `InvalidArgumentException` will be
thrown latter when `$grant->handleResponse()` will be called.

Fix #230

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -205,6 +205,11 @@ abstract class AbstractProvider implements ProviderInterface
         switch ($this->responseType) {
             case 'json':
                 $result = json_decode($response, true);
+
+                if (JSON_ERROR_NONE !== json_last_error()) {
+                    $result = [];
+                }
+
                 break;
             case 'string':
                 parse_str($response, $result);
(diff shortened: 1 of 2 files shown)


ramsey:
Looks good. Thanks!

Issue #213, opened 2015-01-28
Move Providers to their own Repositories


philsturgeon:
- Eventbrite
	- [x] Find a maintainer 
	- [x] Link in the README
	- [x] Remove from league/oauth2-client
- Facebook
	- [x] Make a new GitHub repo
	- [x] Link in the README
	- [x] Remove from league/oauth2-client
	- [x] Document any custom params
- Github
	- [x] Make a new GitHub repo
	- [x] Link in the README
	- [x] Remove from league/oauth2-client
	- [x] Document any custom params
- Google
	- [x] Make a new GitHub repo
	- [x] Link in the README
	- [x] Remove from league/oauth2-client
	- [x] Document any custom params
- Instagram
	- [x] Make a new GitHub repo
	- [x] Link in the README
	- [x] Remove from league/oauth2-client
	- [x] Document any custom params
- LinkedIn
	- [x] Make a new GitHub repo
	- [x] Link in the README
	- [x] Remove from league/oauth2-client
	- [x] Document any custom params
- Microsoft
	- [x] Find a maintainer 
	- [x] Link in the README
	- [x] Remove from league/oauth2-client
- Vkontakte
	- [x] Find a maintainer 
	- [x] Remove from league/oauth2-client
	- [x] Link in the README

You might notice Eventbrite, Microsoft and Vkontakte have been bumped off to their own repos. That's just some advice, but maybe @ramsey you want them to stick around and be official.

SammyK:
If #214 gets approved, this we might want to add, "Create a custom certificate bundle for peer verification," to all the providers in this list. :)

philsturgeon:
Forgot Vkontakte is already done: https://packagist.org/packages/j4k/oauth2-vkontakte

SammyK:
Boom! [Facebook Provider](https://github.com/thephpleague/oauth2-facebook) ready. Not sure how you want to transition removing it from this repo.

philsturgeon:
@SammyK nice! Just whip it out of the 1.0 branch and link it in the README.

SammyK:
I whipped it out...

shadowhand:
I volunteer for Google, but would want it to be under the League's account.

stevenmaguire:
If #242 is merged, we can cross Eventbrite off this punch list.

stevenmaguire:
If #243  is merged, we can cross Microsoft off this punch list.

philsturgeon:
[...]

PR #240, opened 2015-03-20
Adding Uber provider and tests

stevenmaguire:
Uber now has an API that utilizes OAuth2. This proposal includes a provider and supporting tests to offer support for Uber.

### DIFF
--- a/src/Provider/Uber.php
+++ b/src/Provider/Uber.php
@@ -1,0 +1,60 @@
+<?php
+
+namespace League\OAuth2\Client\Provider;
+
+use League\OAuth2\Client\Entity\User;
+use League\OAuth2\Client\Token\AccessToken;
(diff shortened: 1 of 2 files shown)


ramsey:
Thank you for your pull request, but please take note of the README section on [built-in providers](https://github.com/thephpleague/oauth2-client#built-in-providers) and [third-party providers](https://github.com/thephpleague/oauth2-client#third-party-providers), as well as issue #213.

Our plan for v1.0.0 is to move all providers out of this repository into their own separate repositories, which we can link to from the README or a wiki page.

Please create a separate repo for the Uber provider (use one of the existing linked third-party providers as a model, if you like), and then submit a pull request linking to the Uber provider.

Thanks and sorry for the hassle.

PR #247, opened 2015-03-22
Move Google to external package

shadowhand:
As per #213, this moves the Google provider to a [new home](https://github.com/thephpleague/oauth2-google).

Also updates the tests, which were using `Google` instead of a stub provider.


ramsey:
Great stuff. Thanks!

PR #252, opened 2015-03-26
VKontakte email

kudmni:
Some providers (VKontakte) give the email inside access token, so lets take it sends

### DIFF
--- a/src/Token/AccessToken.php
+++ b/src/Token/AccessToken.php
@@ -63,6 +63,11 @@ class AccessToken
             $expiresInFuture = $expires > time();
             $this->expires = $expiresInFuture ? $expires : time() + ((int) $expires);
         }
+        
+        // Some providers (VKontakte) give the email here, so lets take it sends
+		if (!empty($options['email'])) {
+            $this->email = $options['email'];
+        }
     }
 
     /**
--- a/src/Token/AccessToken.php
+++ b/src/Token/AccessToken.php
@@ -65,7 +65,7 @@ class AccessToken
         }
         
         // Some providers (VKontakte) give the email here, so lets take it sends
-		if (!empty($options['email'])) {
+	if (!empty($options['email'])) {
             $this->email = $options['email'];
         }
     }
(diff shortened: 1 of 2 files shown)


ramsey:
Since providers are being removed in v1.0.0, does this make more sense to add to the third-party VKontakte provider repo rather than here? https://github.com/j4k/oauth2-vkontakte

kudmni:
I am already using https://github.com/j4k/oauth2-vkontakte.
However, without this fix access token loses email

ramsey:
I'm really not a fan of adding additional features just to support a single provider. Ideally, we should be flexible enough that these kinds of differences between provider responses can be handled at the provider class level.

I'll do some thinking and looking at the VKontakte provider to see what we need to do in this library to provide that level of flexibility to the down-stream providers.

kudmni:
Alternative is to override method getAccessToken as in my pull request into j4k/oauth2-vkontakte:
https://github.com/j4k/oauth2-vkontakte/pull/3

ramsey:
Yeah. That's a lot of code just to get the email. I'll need to take a look at the VKontakte API to see how they're doing things.

kudmni:
Here is documentation, but it is in russian: https://vk.com/dev/auth_sites
And especially about the email:
"В результате выполнения данного запроса Ваш сервер получит вновь созданный access_token. Вместе с access_token серверу возвращается время жизни ключа expires_in в секундах.
Также сервер передаёт email пользователя, если приложение является сайтом, были запрошены соответствующие права, а также у пользователя есть email, и он разрешил доступ к нему.
Процедуру авторизации приложения необходимо повторять в случае истечения срока действия access_token, смены пользователем своего логина или пароля или удаления приложения из настроек."

In a nutshell, the email is sent together with access token.

ramsey:
Since user data is specific to the individual providers, I've decided to leave that up to each provider to handle. See discussion on #265 for reference—specifically here: https://github.com/thephpleague/oauth2-client/issues/265#issuecomment-94912180

Issue #237, opened 2015-03-13
User Entity cannot be set


marcojanssen:
The User Entity for all the providers are tightly coupled and have no way to be injected or set when retrieving the user details. The only way now is to inherit the provider in our own concrete and override the method, which is not something i prefer.

```
public function userDetails($response, AccessToken $token)
    {
        $user = new User();

        $email = (isset($response->emailAddress)) ? $response->emailAddress : null;
        $location = (isset($response->location->name)) ? $response->location->name : null;
        $description = (isset($response->headline)) ? $response->headline : null;
        $pictureUrl = (isset($response->pictureUrl)) ? $response->pictureUrl : null;

        $user->exchangeArray([
            'uid' => $response->id,
            'name' => $response->firstName.' '.$response->lastName,
            'firstname' => $response->firstName,
            'lastname' => $response->lastName,
            'email' => $email,
            'location' => $location,
            'description' => $description,
            'imageurl' => $pictureUrl,
            'urls' => $response->publicProfileUrl,
        ]);

        return $user;
    }
```

In case of LinkedIn there are more user details we can extract when a r_fullprofile scope is provided.  Are there plans to implement this and if not what are your thoughts about how to implement it? We could add it in the abstract provider but i prefer a seperate object for it.

Will free up some time, if needed, to implement it.

marcojanssen:
As an alternative we can also allow the response or the extra fields to be set in the User entity.

ramsey:
Good recommendation about the User. I'll add this to the 1.0 milestone. If you do submit a pull request for it, be sure to do it against the 1.0 branch.

Thanks!

kevindierkx:
May a suggest the way illuminate/database handles it's [model](https://github.com/illuminate/database/blob/master/Eloquent/Model.php) attributes, it's using a clean way to access dynamic attributes in both an array and object manner. Give me a heads up if you would like me to implement something like it.

ramsey:
How does illuminate/database handle it's model attributes?

kevindierkx:
In a nutshell. During creation of the model using the ```__construct``` or after a query call it utilises the ```fill()``` method provided in the model to fill the ```$attributes``` property. This in turn is used with the ```__get()``` and ```__set()``` methods.

marcojanssen:
Might be worth a try, was thinking of setting a dynamic path and then map the objetc towards that

kevindierkx:
I'll make a fork this evening (GMT+0001). ;)

[...]

PR #258, opened 2015-04-09
Fixed headers being parsed incorrectly

kevindierkx:
As mentioned in #237 authentication headers would be wrapped in a 'headers' header.


kevindierkx:
@ramsey Hey, could you merge this? It's a requirement for #237. Thanks in advance ;)

ramsey:
Thanks for pinging me. I don't think GitHub sent me an email about this.

PR #262, opened 2015-04-18
send  param in redirectHandler

xjchengo:
add a param to redirectHandler, so we can easily get state and so on in redirectHandler.

```php
$provider->setRedirectHandler(function($url, $provider){
    $_SESSION['oauth2state'] = $provider->state;
    header('Location: '.$url);
    exit;
}
```


ramsey:
Thank you for your contribution!

PR #263, opened 2015-04-18
Exception handle

xjchengo:
add a errorChecker property in provider, so we can handle custom error.
When I wrote a third party provider, i found it return '{"errcode": xxxx, "errmsg": "invalid token"}' when encountering error. It is hard for oauth2-client 1.0 to handle this error type.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -43,6 +43,8 @@ abstract class AbstractProvider implements ProviderInterface
 
     protected $redirectHandler;
 
+    protected $errorChecker;
+
     /**
      * @var int This represents: PHP_QUERY_RFC1738, which is the default value for php 5.4
      *          and the default encryption type for the http_build_query setup
@@ -58,6 +60,14 @@ abstract class AbstractProvider implements ProviderInterface
         }
 
         $this->setHttpClient($httpClient ?: new CurlHttpAdapter());
+
+        if (!isset($options['errorChecker'])) {
+            $this->setErrorChecker(function ($result) {
+                if (isset($result['error']) && ! empty($result['error'])) {
+                    throw new IDPException($result);
+                }
+            });
+        }
     }
 
     public function setHttpClient(HttpAdapterInterface $client)
@@ -217,12 +227,8 @@ abstract class AbstractProvider implements ProviderInterface
 
                 break;
         }
-
-        if (isset($result['error']) && ! empty($result['error'])) {
-            // @codeCoverageIgnoreStart
-            throw new IDPException($result);
-            // @codeCoverageIgnoreEnd
-        }
+        $checker = $this->errorChecker;
+        $checker($result);
 
         $result = $this->prepareAccessTokenResult($result);
 
@@ -376,5 +382,10 @@ abstract class AbstractProvider implements ProviderInterface
     public function setRedirectHandler(Closure $handler)
     {
         $this->redirectHandler = $handler;
+    }
+
+    public function setErrorChecker(Closure $checker)
+    {
+        $this->errorChecker = $checker;
     }
 }
--- a/src/Exception/IDPException.php
+++ b/src/Exception/IDPException.php
@@ -6,20 +6,24 @@ class IDPException extends \Exception
 {
     protected $result;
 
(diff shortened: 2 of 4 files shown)


src/Provider/AbstractProvider.php
```diff
@@ -217,12 +227,8 @@ public function getAccessToken($grant = 'authorization_code', $params = [])
...
                 break;
         }
-
-        if (isset($result['error']) && ! empty($result['error'])) {
-            // @codeCoverageIgnoreStart
-            throw new IDPException($result);
-            // @codeCoverageIgnoreEnd
-        }
+        $checker = $this->errorChecker;
+        $checker($result);
```
ramsey: I guess this means that errorChecker must always be a callable? We should probably do a check for that somewhere (the constructor?) and throw an exception if it's not the case. In v1.0, as I said on your other pull request, I'd like to change these properties to protected and have real getters/setters on them, so that we can type-hint for things like `callable`.

xjchengo: Yes, errorChecker must be callable. the setter of errorChecker has type-hint of Closure. After rethinking, i think my code is terrible. thephpleague/oauth2-client  v1 is oriented to third party provider authors and should easy for these authors to extend. Maybe add a abstract method errorChecker in  AbstractProvider is better and enough. So third party provider authors can define what is error according to third party provider document.

xjchengo:
I think current IDPException constructor is lack of adaptability. If the third party provider error response don't put error message into 'error' or 'message' and error code into 'code', we have to transform error response before throwing the exception. Why not add $message and $code as required params to IDPException constructor?  Coordinate with above-mentioned abstract errorChecker method, error check and throw is easier.

src/Provider/AbstractProvider.php
xjchengo: And I don't catch what the real getter/setter you mean is. Can you explain it please?

ramsey:
Did you happen to see @SammyK's pull request #266? It seems related and might provide the functionality you need by allowing third-party providers to parse the exception first before throwing it.

SammyK:
I like that @xjchengo's solution goes a step further and extracts the `if (isset($result['error']) && ! empty($result['error'])) {}` check out of the method. That's crucial when the error response doesn't have an `error` key. It might be beneficial to pass the whole request object so you can check for the HTTP response codes in the event the body is empty on error. I can refactor my PR to do this if you like.

ramsey:
Yeah. I see no way we can accommodate all the potential responses each provider might give us. Rather than accepting an array, we should accept specific scalar values (like `$message` and `$code` here in @xjchengo's PR) and let the providers parse it before giving it to the exception.

SammyK:
:+1: @xjchengo do you want to take that on or do you want me to?

ramsey:
Let's change the name of this class, too. `IDPException` has always been so confusing to me. It would be better named as `IdentityProviderException` to avoid the confusion and so we don't accidentally overload it with other meaning.

Perhaps move it to the Provider namespace as `Provider\Exception\IdentityProviderException`?

SammyK:
:+1: for rename & namespace move.

Should each provider's unique exception e.g. `FacebookProviderException` extend from this one? That way you can always catch just an `IdentityProviderException` and it will also catch any provider-specific exceptions.

ramsey:
Yes. That's a good use of this. I'm not exactly sure how to enforce that, though.

xjchengo:
@SammyK i will take it on. 
@ramsey Yes, after i send this PR, i see @SammyK 's PR. The getType method in IDPExcepton is strange. In my opinion, end user seldom use this method? They can use getMessage. Can we remove the getType method? 
If we move IdentityProviderException to the Provider namespace, move Exception folder to Provider folder or just add a classmap autoload in composer.json?

ramsey:
I'm not sure why we would need to use the classmap autoloader. Can you elaborate on what you mean?

xjchengo:
If we don't move the Exception folder and just change the namespace of the exception, it violates psr-4.

ramsey:
Yes. Please move it. That's what I meant. :-)

ramsey:
Closing this, as it was handled in #269.

PR #264, opened 2015-04-20
add array access to User Entity

xjchengo:
make User more extendable and easier to use

### DIFF
--- a/src/Entity/User.php
+++ b/src/Entity/User.php
@@ -2,116 +2,72 @@
 
 namespace League\OAuth2\Client\Entity;
 
-class User
+use ArrayAccess;
+use OutOfRangeException;
+
+class User implements ArrayAccess
 {
-    protected $uid;
-    protected $nickname;
-    protected $name;
-    protected $firstName;
-    protected $lastName;
-    protected $email;
-    protected $location;
-    protected $description;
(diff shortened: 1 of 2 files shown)


ramsey:
I'm not a fan of magic method properties, as you can't document them, and they also become problematic in testing and code coverage. One of my goals for 1.0 is to convert these properties into protected members and then use real getters and setters with proper docblocks (as is [the League way](http://thephpleague.com/#quality)).

kevindierkx:
@ramsey As discussed in #237 the "real" getters and setters are not always the same for user responses. A class like this can dynamically manage these differences. Or we should provide a User entity in the provider to make it customisable.

ramsey:
@kevindierkx I understand what you're saying. I think the problem here is that OAuth 2 doesn't define anything to do with a User entity, so this is even sort of out of scope for this library. As a result, each provider handles responses a little bit differently and a little bit off-spec. I think you're right that we should provide some base User interface or abstract User and let each provider handle it. I'll write up a recommendation for 1.0 and put it in the Issues section.

ramsey:
I removed the User entity in #267 and replaced it with `Provider\UserInterface` that all third-party providers should use.

PR #266, opened 2015-04-20
Add feature for provider to parse error response before throwing IDPException

SammyK:
Facebook error responses are returned like so:

    {"error":{"message":"Some auth error","type":"OAuthException","code":191}}

So when there is an error and it gets to the `IDPException` constructor, it will try to send an array to the `$exception` argument of `parent::__construct()` which throws the following error:

    Fatal error: Wrong parameters for Exception([string $exception [, long $code [, Exception $previous = NULL]]]) in %s on line %d

This PR adds a new `throwIDPException()` method to the `AbstractProvider` so that a provider can parse the error response before the `IDPException` is thrown.

In the Facebook provider case it would look like this:

```php
class FacebookProvider extends AbstractProvider
{
    public function throwIDPException(array $result)
    {
        $result = $result['error'];

        throw new IDPException($result);
    }
}
```

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -223,7 +223,7 @@ abstract class AbstractProvider implements ProviderInterface
 
         if (isset($result['error']) && ! empty($result['error'])) {
             // @codeCoverageIgnoreStart
-            throw new IDPException($result);
+            $this->throwIDPException($result);
             // @codeCoverageIgnoreEnd
         }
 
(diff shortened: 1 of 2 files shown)


ramsey:
Looks good. I think this may help to solve some other issues that people have been reporting. Thanks!

ramsey:
Hmm. It looks like you made the branch on this repo; I didn't realize you had access. :-)

Is it okay if I delete the branch?

SammyK:
Yeah, sorry about that. I didn't realize that until I pushed it up. Lol. Yeah go ahead and delete the branch. And feel free to remove me from direct access to this repo. I don't want to `git push origin master -f` on accident. lol. :)

ramsey:
@SammyK Looks like you're in the oauth2-client team, which has access to all the oauth2-* repos. I don't have access to manage that team, and it's not a big deal. Just please don't `--force` on anything. :-)

SammyK:
Ha! No worries. From now I'm I'm just going to send PR's from my repo and remove any `origin` remotes to this one. :)

PR #267, opened 2015-04-21
UserInterface Contract for Third-party Providers

### LINKED ISSUE #265: [RFC] Provider User Entity Interface/Abstract
OAuth 2.0 does not define a User entity. However, we're providing an `Entity\User` class in this library. As a result of how various providers handle users and user data, we've had to modify this class numerous times to support each provider. Now that there are no more providers in version 1.0, it no longer makes sense to support a User entity that works for all the providers.

My recommendation is to remove `Entity\User` in v1.0. In its place, we should provide a basic `Provider\UserInterface` and possibly an abstract `Provider\User` with bare-bones functionality. Each provider will need to upgrade to use at least the interface by the v1.0 release. The management of user entities will now be the responsibility of each provider.

We should get feedback from each of the third-party providers for what the `Provider\UserInterface` should include.

Refer to #237 and #264 for more discussion on this topic.

ramsey:
As the user entity is not defined by OAuth2 and varies widely among providers, I feel it is necessary to reduce our support for maintaining such an entity. Instead, we should provide a bare-minimum contract for all providers to implement, and each provider repository may handle the individual implementation details necessary to provide full functionality for the service they are integrating with.

This pull request is still open for discussion (see #265) and is provided here for illustration and discussion purposes. Once the third-party providers agree on this interface, then I'll merge it in to the upstream repository.

When merged, this pull request closes #265.


test/src/Provider/Fake.php
```diff
@@ -28,3 +28,11 @@ public function userDetails($response, AccessToken $token)
         return new User;
     }
 }
+
+class User implements UserInterface
```
shadowhand: I think this should be defined in `test/src/Provider/Fake/User.php`

shadowhand:
I think there is some additional clean up work to be done in the `ProviderInterface`, because some methods such as `getUserEmail(AccessToken $token)` should no longer exist if the `UserInterface` does not support them.

PR #269, opened 2015-04-22
rename IDPException

xjchengo:
The getType method in IdentityProviderException is strange. Maybe we can remove it?


src/Provider/AbstractProvider.php
```diff
@@ -307,16 +305,13 @@ public function userScreenName($response, AccessToken $token)
...
      * @param array $result
      *
-     * @throws IDPException
+     * @throws IdentityProviderException
      */
-    public function throwIDPException(array $result)
-    {
-        throw new IDPException($result);
-    }
+    abstract public function errorCheck(array $result);
```
ramsey: Should this be abstract, or should we provide a default behavior?

src/Provider/Exception/IdentityProviderException.php
```diff
@@ -0,0 +1,43 @@
...
+
+namespace League\OAuth2\Client\Provider\Exception;
+
+class IdentityProviderException extends \Exception
+{
+    protected $result;
+
+    public function __construct($message, $code, $result)
+    {
+        $this->result = $result;
```
ramsey: What exactly is result going to be? An array, an object? Is it just a raw value we're allowing the provider to pass around so the user can access it on the exception?

ramsey:
Looks good. I just had a couple of questions, but otherwise, I think this does the job. Thanks!

src/Provider/AbstractProvider.php
xjchengo: I have used two oauth2 service and they all have their custom error response. So maybe it is hard to provider a default errorCheck suitable for most cases. And more important, if we provider a default behavior, third party provider authors will tend to be less concerned about error check.  Once the default behavior isn't suitable and the author overlooks it, the end user will be harmed. So in my opinion, being abstract is better. It enforces paying attention to error check.

src/Provider/Exception/IdentityProviderException.php
xjchengo: $result is same with argument passed to errorCheck. So it is an array. Sometimes the error response have information other than error message and code. So we need make the user can access it on the exception.

PR #271, opened 2015-04-25
Add new getResponseErrorMessage method

shadowhand:
- add `prepareResponse` method to parse responses
- use `prepareResponse` within `fetchProviderData`
- handle wrapped errors in `IDPException`

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -202,19 +202,7 @@ abstract class AbstractProvider implements ProviderInterface
             // @codeCoverageIgnoreEnd
         }
 
-        switch ($this->responseType) {
-            case 'json':
-                $result = json_decode($response, true);
-
-                if (JSON_ERROR_NONE !== json_last_error()) {
-                    $result = [];
-                }
-
-                break;
-            case 'string':
-                parse_str($response, $result);
-                break;
-        }
+        $result = $this->prepareResponse($response);
 
         if (isset($result['error']) && ! empty($result['error'])) {
             // @codeCoverageIgnoreStart
@@ -225,6 +213,34 @@ abstract class AbstractProvider implements ProviderInterface
         $result = $this->prepareAccessTokenResult($result);
 
         return $grant->handleResponse($result);
+    }
+
+    /**
+     * Prepare the response, parsing according to configuration and returning
+     * the response as an array.
+     *
+     * @param  string $response
+     * @return array
+     */
+    protected function prepareResponse($response)
+    {
+        $result = [];
+
+        switch ($this->responseType) {
+            case 'json':
+                $json = json_decode($response, true);
+
+                if (JSON_ERROR_NONE === json_last_error()) {
+                    $result = $json;
+                }
+
+                break;
+            case 'string':
+                parse_str($response, $result);
+                break;
+        }
+
+        return $result;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


shadowhand:
@ramsey I think this is a pretty useful addition. Without it, all the errors coming from Google end up as `Unknown Error`

src/Provider/AbstractProvider.php
```diff
@@ -349,14 +349,20 @@ protected function fetchProviderData($url, array $headers = [])
             $response = $request->getBody();
         } catch (BadResponseException $e) {
             // @codeCoverageIgnoreStart
-            $raw_response = explode("\n", $e->getResponse());
-            throw new IDPException(end($raw_response));
+            $message = $this->getResponseErrorMessage($e->getResponse()->getBody());
+            throw new IDPException($message);
```
shadowhand: Actually, on second look... I'm not sure how `IDPException` is ever supposed to work correctly as written, because this code is assuming that the response is plaintext, but [IDPException is expecting an array](https://github.com/thephpleague/oauth2-client/blob/9f2725a336d3aec025b578204938db934a12a25d/src/Exception/IDPException.php#L13-L23).

ramsey:
Are you intending to merge this into master? In the 1.0 branch, `Exception\IDPException` has moved to `Provider\Exception\IdentityProviderException`. If you do intend to merge this into master, I'll go ahead and merge it, but it won't cleanly merge up into 1.0, so you might want to review that and do another PR there, if applicable.

shadowhand:
@ramsey yes it should go into master. (I don't have the ability to merge.) I can do a separate PR for 1.0 branch once we decide on a fix here.

src/Provider/AbstractProvider.php
ramsey: Yeah. `IDPException` in master is expecting an array, and then it attempts to parse values out of that array. I don't think I can merge this into master, as a result, since it will break other providers' implementations.

Thoughts?

shadowhand:
@ramsey completely changed the PR and description. I think this solves the issue much more elegantly and doesn't modify anything for existing implementations.

src/Provider/AbstractProvider.php
```diff
@@ -202,12 +202,36 @@ public function getAccessToken($grant = 'authorization_code', $params = [])
...
+    }
+
+    /**
+     * Prepare the response, parsing according to configuration and returning
+     * the response as an array.
+     *
+     * @param  string $response
+     * @return array
+     */
+    protected function prepareResponse($response)
```
shadowhand: Moved the response parsing to a separate method so that `fetchProviderData` could make use of it.

test/src/Exception/IDPExceptionTest.php
```diff
@@ -42,6 +42,13 @@ public function testAsStringWithCode()
         $this->assertEquals('message: 404: message', (string)$exception);
     }
 
+    public function testAsWrapped()
+    {
+        $exception = new IDPException(array('error' => array('error' => 'message')));
```
shadowhand: Additional test for the new wrapped support.

ramsey:
I like it! Thanks!

shadowhand:
@ramsey if this could be tagged for release, it would be great!

ramsey:
Done! 0.11.0

shadowhand:
Awesome, thanks!

PR #273, opened 2015-04-26
Fully document the ProviderInterface

shadowhand:
Move existing documentation from `AbstractProvider` and fill in any missing docblocks.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -114,6 +114,16 @@ abstract class AbstractProvider implements ProviderInterface
         return $client;
     }
 
+    // Implementing these interfaces methods should not be required, but not
+    // doing so will break HHVM because of https://github.com/facebook/hhvm/issues/5170
+    // Once HHVM is working, delete the following abstract methods.
+    abstract public function urlAuthorize();
+    abstract public function urlAccessToken();
+    abstract public function urlUserDetails(AccessToken $token);
+    abstract public function userDetails($response, AccessToken $token);
+    abstract public function errorCheck(array $result);
+    // End of methods to delete.
+
     public function getScopes()
     {
         return $this->scopes;
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -8,33 +8,69 @@ use Ivory\HttpAdapter\HttpAdapterException;
 use Ivory\HttpAdapter\HttpAdapterInterface;
 use League\OAuth2\Client\Provider\Exception\IdentityProviderException;
 use League\OAuth2\Client\Grant\GrantInterface;
-use League\OAuth2\Client\Token\AccessToken as AccessToken;
+use League\OAuth2\Client\Token\AccessToken;
 use UnexpectedValueException;
 
 abstract class AbstractProvider implements ProviderInterface
 {
+    /**
+     * @var string
+     */
     public $clientId = '';
 
+    /**
+     * @var string
+     */
     public $clientSecret = '';
 
+    /**
+     * @var string
+     */
     public $redirectUri = '';
 
+    /**
+     * @var string
+     */
     public $state;
 
+    /**
+     * @var string
+     */
     public $name;
 
+    /**
+     * @var string
+     */
     public $uidKey = 'uid';
 
+    /**
+     * @var array
+     */
     public $scopes = [];
 
+    /**
+     * @var string
+     */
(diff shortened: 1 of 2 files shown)


src/Provider/ProviderInterface.php
```diff
@@ -2,35 +2,143 @@
...
     public function getHeaders($token = null);
 
+    /**
+     * Throws an IdentityProviderException when an error response is received
+     *
+     * @throws IdentityProviderException
+     * @param array $result
+     * @return void
+     */
+    public function errorCheck(array $result);
```
shadowhand: @ramsey This method was missing from the interface. As per our discussion earlier about the error handling flow, I plan to modify this method (and name it better).

src/Provider/AbstractProvider.php
```diff
@@ -74,43 +74,21 @@ public function getHttpClient()
         return $client;
     }
 
-    /**
-     * Get the URL that this provider uses to begin authorization.
-     *
-     * @return string
-     */
+    // ProviderInterface
```
ramsey: Since these methods are in the interface, they shouldn't be in the abstract, unless we're actually implementing something, right? Let's remove them, unless you think there's a reason to keep them here.

Additionally, if are implementing them, we should use `{@inheritDoc}` in docblocks instead of this style of comment:

```
/**
 * {@inheritDoc}
 */
```

shadowhand: That's a common misconception, and not what `inherticdoc` is meant for. As per [the documentation](http://www.phpdoc.org/docs/latest/guides/inheritance.html):

> Currently some applications have DocBlocks containing just the {@inheritDoc} inline tag to indicate that their complete contents should be inherited. This usage breaks with the PHPDoc Standard as summaries cannot contain inline tags and inheritance is automatic; you do not need to define a special tag for it.

shadowhand: > Since these methods are in the interface, they shouldn't be in the abstract

Removing them does work, I'll update the PR.

ramsey: Good to know. Thanks!

shadowhand:
@ramsey I'm not sure why this build failed only on HHVM... any clue?

ramsey:
The failure message on HHVM is:

```
Fatal error: Class Mock_AbstractProvider_c5873f51 contains abstract method (urlAuthorize) and must therefore be declared abstract or implement the remaining methods in /home/travis/build/thephpleague/oauth2-client/vendor/phpunit/phpunit-mock-objects/src/Framework/MockObject/Generator.php(301) : eval()'d code on line 1
```

I'm not sure why it's specific to HHVM, though.

shadowhand:
@ramsey found the issue, it's a HHVM bug: sebastianbergmann/phpunit-mock-objects#223 and facebook/hhvm#5170. Should I apply the work around (define the methods in the abstract class) or are we okay with breaking HHVM usage?

ramsey:
Is this only because our mock provider doesn't fully implement all the methods? Or is it a general issue with abstract classes in HHVM, where they always need to implement the full interface?

If it's only an issue with our mock provider for the tests, then please define the methods in the mock object for the test. Otherwise, let's think about it for a bit. I want to follow up with HHVM folks to see if this is something they'll be fixing in future versions.

shadowhand:
@ramsey based on the HHVM issue I referenced, it looks like they intend to fix it, but have no timeline for doing so. I think the best thing to do for now would be to implement all the methods in the abstract.

ramsey:
Okay. That's fine. When HHVM does fix it, we can remove them.

src/Provider/AbstractProvider.php
```diff
@@ -75,42 +114,13 @@ public function getHttpClient()
...
     }
 
-    /**
-     * Get the URL that this provider uses to begin authorization.
-     *
-     * @return string
-     */
+    // Implementing these interfaces methods should not be required, but not
+    // doing so will break HHVM because of https://github.com/facebook/hhvm/issues/5170
+    // Once HHVM is working, delete the following abstract methods.
```
shadowhand: @ramsey added here as a separate commit that can be reverted in the future.

shadowhand:
Strange the coverage decreased, since no actual code was modified. I guess it's because of the added abstract methods.

Also, I [made a quick script](https://gist.github.com/shadowhand/915ff25729b038577475) to check these after I kept missing methods.

ramsey:
Thanks!

PR #274, opened 2015-04-26
Add consistent response handling

shadowhand:
- add `parseResponse` method
- use `parseResponse` with `getAccessToken`
- use `parseResponse` with `fetchProviderData`
- success and error response handling use the same path
- update `getUserFoo` methods accordingly

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -202,9 +202,7 @@ abstract class AbstractProvider implements ProviderInterface
                 // @codeCoverageIgnoreEnd
             }
         } catch (HttpAdapterException $e) {
-            // @codeCoverageIgnoreStart
             $response = (string) $e->getResponse()->getBody();
-            // @codeCoverageIgnoreEnd
         }
 
         $result = $this->parseResponse($response);
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -206,12 +206,34 @@ abstract class AbstractProvider implements ProviderInterface
             // @codeCoverageIgnoreEnd
         }
 
+        $result = $this->parseResponse($response);
+
+        // @codeCoverageIgnoreStart
+        $this->errorCheck($result);
+        // @codeCoverageIgnoreEnd
+
+        $result = $this->prepareAccessTokenResult($result);
+
+        return $grant->handleResponse($result);
+    }
+
+
+    /**
+     * Parse the response, according to the provider response type.
+     *
+     * @param  string $response
+     * @return array
+     */
+    protected function parseResponse($response)
+    {
+        $result = [];
+
         switch ($this->responseType) {
             case 'json':
-                $result = json_decode($response, true);
+                $json = json_decode($response, true);
 
-                if (JSON_ERROR_NONE !== json_last_error()) {
-                    $result = [];
+                if (JSON_ERROR_NONE === json_last_error()) {
+                    $result = $json;
                 }
 
                 break;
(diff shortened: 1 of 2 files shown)


shadowhand:
@ramsey as per our discussion earlier today, which I think has (sadly) been rebased out of existence.

src/Provider/AbstractProvider.php
```diff
@@ -357,12 +379,17 @@ protected function fetchProviderData($url, array $headers = [])
...
         } catch (HttpAdapterException $e) {
             // @codeCoverageIgnoreStart
-            $raw_response = explode("\n", (string) $e->getResponse()->getBody());
-            throw new IDPException(end($raw_response));
+            $response = (string) $e->getResponse()->getBody();
             // @codeCoverageIgnoreEnd
         }
 
-        return $response;
+        $result = $this->parseResponse($response);
```
shadowhand: @ramsey this is what I was talking about WRT changing the response type of `fetchProviderData`. Tests updated accordingly, but see my comments on #267 about removing the `getUserFoo` methods due to the limited `UserInterface`.

src/Provider/AbstractProvider.php
```diff
@@ -206,12 +206,34 @@ public function getAccessToken($grant = 'authorization_code', $params = [])
...
+
         switch ($this->responseType) {
             case 'json':
-                $result = json_decode($response, true);
+                $json = json_decode($response, true);
 
-                if (JSON_ERROR_NONE !== json_last_error()) {
-                    $result = [];
+                if (JSON_ERROR_NONE === json_last_error()) {
+                    $result = $json;
```
ramsey: If there is a JSON error, is it okay for `parseResponse()` to return an empty array? That's what it looks like it would do. Should it instead throw an exception, since it couldn't parse the response?

src/Provider/AbstractProvider.php
ramsey: I agree. Let's do that. It doesn't make sense for the provider to have all these helper methods. Some providers will use them, others won't. We should leave it up to them to return a `UserInterface`, and they can extend it however it seems fit for the service they are using.

src/Provider/AbstractProvider.php
shadowhand: Probably a good idea.

src/Provider/AbstractProvider.php
```diff
@@ -201,17 +202,37 @@ public function getAccessToken($grant = 'authorization_code', $params = [])
...
+    {
+        $result = [];
+
         switch ($this->responseType) {
             case 'json':
                 $result = json_decode($response, true);
 
                 if (JSON_ERROR_NONE !== json_last_error()) {
-                    $result = [];
+                    throw new UnexpectedValueException('Unable to parse client response');
```
shadowhand: @ramsey exception for failure to parse JSON.

test/src/Provider/AbstractProviderTest.php
```diff
@@ -212,6 +215,60 @@ public function testErrorResponsesCanBeCustomizedAtTheProvider()
...
+          'clientSecret' => 'mock_secret',
+          'redirectUri' => 'none',
+        ]);
+
+        $response = m::mock('Ivory\HttpAdapter\Message\ResponseInterface');
+        $response->shouldReceive('getBody')
+                 ->times(1)
+                 ->andReturn('{"error":"BadResponse","code":500}');
+
+        $exception = new HttpAdapterException();
```
shadowhand: @ramsey is this a good test? Is it okay to be testing vendor exceptions, or was that why the code coverage was being ignored?

ramsey:
Thanks! Merging in.

test/src/Provider/AbstractProviderTest.php
ramsey: This test looks good. Frankly, I'm not sure why the code coverage was being ignored all over the place. That's been on my list of things to clean up for v1.0.

PR #275, opened 2015-04-27
Add methods for creating and sending requests

### LINKED ISSUE #268: [RFC] ProviderInterface::getAuthenticatedRequest() method
This would give users of Providers a method that returns an already-authenticated HTTP client, using the token retrieved in the OAuth 2.0 flow. They may then use this client object to make authenticated requests of whatever provider API they are using, without the OAuth 2.0 Provider classes becoming API wrappers.

From @shadowhand in the comments on https://github.com/thephpleague/oauth2-client/issues/265#issuecomment-94943566:

> One of the things that always bothered me about this library (and am moving towards solving with #232) is that it doesn't do very much to make **using** the provider APIs easier, it only gets us to the point of having an authenticated token.
> 
> I'll be very clear here and say that **the goal is not to create API wrappers** with providers. However, I think a very valuable part of this package is that it makes it **easier to use provider APIs** and so any changes we are proposing here should keep that in mind.
> 
> To this end, what are valuable things this package can do to make API usage easier?
> 
> 1. Ensure that providers contain a method to fetch an API base URL. This is not very useful by itself, but it makes writing multiple API calls much DRYer.
> 2. Ensure that providers contain a method to add authentication to a request. This is what #232 did (see also thephpleague/oauth1-client#18) and it immediately improved the code I was working on that made API calls.
> 3. Provide ways to integrate more cleanly with dependent packages. The client already uses Guzzle, so it should expose methods that work with those classes. For instance, having a method like this would save a lot of repeated code:
> 
> ```php
> /**
>  * Create an authenticated client to access an API URL.
>  *
>  * @param  string $method
>  * @param  string $path
>  * @param  AccessToken $token
>  * @return Psr\Http\Message\RequestInterface
>  */
> public function getAuthenticatedRequest($method, $path, AccessToken $token);
> ```
> 
> This would allow us to write code like:
> 
> ```php
> $httpRequest = $oauthProvider->getAuthenticatedRequest('GET', 'me?fields=id,name,email', $token);
> $httpResponse = $httpClient->send($httpRequest);
> $me = json_decode($httpResponse->getBody());
> ```
> 
> Personally I would find this to be a *very* useful and complete example of how to build upon what the package provides and represents best practice.

shadowhand:
- add `getAuthenticatedRequest` method to get a request builder
- add `getResponse` method to send requests and parse the response
- use new methods for handling user detail calls

Fixes #268

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -6,6 +6,7 @@ use Closure;
 use Ivory\HttpAdapter\CurlHttpAdapter;
 use Ivory\HttpAdapter\HttpAdapterException;
 use Ivory\HttpAdapter\HttpAdapterInterface;
+use Ivory\HttpAdapter\Message\RequestInterface;
 use League\OAuth2\Client\Provider\Exception\IdentityProviderException;
 use League\OAuth2\Client\Grant\GrantInterface;
 use League\OAuth2\Client\Token\AccessToken as AccessToken;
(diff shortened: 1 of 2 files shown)


src/Provider/AbstractProvider.php
```diff
@@ -216,6 +218,54 @@ public function getAccessToken($grant = 'authorization_code', $params = [])
...
+     * Often used to create calls against an API that requires authentication.
+     *
+     * @param  string $method
+     * @param  string $url
+     * @param  AccessToken $token
+     * @return RequestInterface
+     */
+    public function getAuthenticatedRequest($method, $url, AccessToken $token)
+    {
+        $request = new Request($url, $method);
```
ramsey: Should this call `getHttpClient()` to build the request, so that the user may choose whatever HTTP client they wish to use?

src/Provider/AbstractProvider.php
```diff
@@ -216,6 +217,58 @@ public function getAccessToken($grant = 'authorization_code', $params = [])
...
+     * @param  AccessToken $token
+     * @return RequestInterface
+     */
+    public function getAuthenticatedRequest($method, $url, AccessToken $token)
+    {
+        $factory = $this->getHttpClient()
+            ->getConfiguration()
+            ->getMessageFactory();
+
+        $request = $factory->createRequest($url, $method);
```
shadowhand: @ramsey modified to use the client message factory. Good call.

ramsey:
Thanks! Looks good.

PR #276, opened 2015-04-28
Modify how provider response errors are handled

shadowhand:
- change `public checkError` into `protected checkResponse`
- remove custom `toString` handling in `IdentityProviderException`
- clean up tests

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -163,7 +163,6 @@ abstract class AbstractProvider implements ProviderInterface
     abstract public function urlAccessToken();
     abstract public function urlUserDetails(AccessToken $token);
     abstract public function userDetails($response, AccessToken $token);
-    abstract public function errorCheck(array $result);
     // End of methods to delete.
 
     public function getScopes()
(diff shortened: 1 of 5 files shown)


shadowhand:
@ramsey after looking through the commits related to when `checkError` was introduced, I decided that using throwing the exception from the provider does make sense. What didn't make sense is that it was exposed as part of the public API, especially in light of the functionality in #275 for handling request/response cycle through the provider.

End result is mostly clean up of tests and making things more consistent internally.

src/Provider/AbstractProvider.php
```diff
@@ -295,22 +286,30 @@ protected function parseResponse($response)
...
     }
 
     /**
+     * Check a provider response for errors.
+     *
+     * @throws IdentityProviderException
+     * @param  array $response
+     * @return void
+     */
+    abstract protected function checkResponse(array $response);
```
shadowhand: I don't see any reason for this to be part of the `ProviderInterface`, because it is a self-referencing method... it doesn't make sense to use it externally.

src/Provider/Exception/IdentityProviderException.php
```diff
@@ -4,40 +4,20 @@
...
-
-        return $result;
-    }
-
-    /**
-     * To make debugging easier.
-     *
-     * @return string The string representation of the error.
-     */
-    public function __toString()
```
shadowhand: No reason to include this anymore, since `getType` is no longer customizable.

test/src/Provider/AbstractProviderTest.php
```diff
@@ -302,33 +296,3 @@ public function testAuthenticatedRequestAndResponse()
         $this->assertSame(['example' => 'response'], $result);
     }
 }
-
-class MockProvider extends \League\OAuth2\Client\Provider\AbstractProvider
```
shadowhand: No need to have this class, we already have the `Fake` provider.

shadowhand:
@ramsey this one is ready to go when you have time to look at it.

ramsey:
Finally got a chance to review your changes and read your notes. Looks great. Thanks!

PR #277, opened 2015-04-28
Generate cryptographically secure states

### LINKED ISSUE #212: [Proposal] Add CSPRNG's to generate CSRF value
Hey guys!

The [code that generates the CSRF value](https://github.com/thephpleague/oauth2-client/blob/master/src/Provider/AbstractProvider.php#L123) makes use of `uniqid()` and `rand()` which do not generate **cryptographically secure** pseudo-random values. Since the CSRF (`state`) value is a security feature to guard against [forgery attacks](http://en.wikipedia.org/wiki/Cross-site_request_forgery), it is paramount that the CSRF value is not forgeable (and it currently is.) In order to make this value forge-proof, we need to generate a pseudo-random value that is [cryptographically secure](http://en.wikipedia.org/wiki/Cryptographically_secure_pseudorandom_number_generator).

PHP particularly sucks at providing CSPRNG's, but there are a few options like [`openssl_random_pseudo_bytes()`](http://php.net/openssl_random_pseudo_bytes) and [`mcrypt_create_iv()`](http://php.net/mcrypt_create_iv). Unfortunately system support for these functions varies.

We recently added [injectable CSPRNG's](https://github.com/facebook/facebook-php-sdk-v4/tree/master/src/Facebook/PseudoRandomString) into the Facebook PHP SDK v4.1. Would you like to implement CSPRNG's using a similar method? Or perhaps we could make use of [Anthony Ferrara](https://twitter.com/ircmaxell)'s excellent [RandomLib](https://github.com/ircmaxell/RandomLib) package?

What do you guys think? :)

shadowhand:
- modify constructor to take N collaborators
- use [ircmaxell/random-lib](https://github.com/ircmaxell/RandomLib) for state generation
- update tests

Fixes #212

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -10,6 +10,7 @@ use Ivory\HttpAdapter\Message\RequestInterface;
 use League\OAuth2\Client\Provider\Exception\IdentityProviderException;
 use League\OAuth2\Client\Grant\GrantInterface;
 use League\OAuth2\Client\Token\AccessToken;
+use RandomLib\Factory as RandomFactory;
 use UnexpectedValueException;
 
 abstract class AbstractProvider implements ProviderInterface
@@ -80,6 +81,11 @@ abstract class AbstractProvider implements ProviderInterface
     protected $httpClient;
 
     /**
+     * @var RandomFactory
+     */
+    protected $randomFactory;
+
+    /**
      * @var Closure
      */
     protected $redirectHandler;
@@ -106,6 +112,11 @@ abstract class AbstractProvider implements ProviderInterface
             $collaborators['httpClient'] = new CurlHttpAdapter();
         }
         $this->setHttpClient($collaborators['httpClient']);
+
+        if (empty($collaborators['randomFactory'])) {
+            $collaborators['randomFactory'] = new RandomFactory();
+        }
+        $this->setRandomFactory($collaborators['randomFactory']);
     }
 
     public function setHttpClient(HttpAdapterInterface $client)
@@ -120,6 +131,29 @@ abstract class AbstractProvider implements ProviderInterface
         $client = $this->httpClient;
 
         return $client;
+    }
+
+    /**
+     * Set the instance of the CSPRNG random generator factory to use.
+     *
+     * @param  RandomFactory $factory
+     * @return $this
+     */
+    public function setRandomFactory(RandomFactory $factory)
+    {
+        $this->randomFactory = $factory;
+
+        return $this;
+    }
+
+    /**
+     * Get the instance of the CSPRNG random generatory factory.
+     *
+     * @return RandomFactory
+     */
+    public function getRandomFactory()
+    {
+        return $this->randomFactory;
     }
 
     // Implementing these interfaces methods should not be required, but not
@@ -142,17 +176,46 @@ abstract class AbstractProvider implements ProviderInterface
         $this->scopes = $scopes;
     }
 
+    /**
+     * Get a new random string to use for auth state.
+     *
+     * @param  integer $length
+     * @return string
+     */
+    protected function getRandomState($length = 32)
+    {
+        $generator = $this
+            ->getRandomFactory()
+            ->getMediumStrengthGenerator();
+
+        return $generator->generateString($length);
+    }
+
     public function getAuthorizationUrl(array $options = [])
     {
-        $this->state = isset($options['state']) ? $options['state'] : md5(uniqid(rand(), true));
+        if (empty($options['state'])) {
+            $options['state'] = $this->getRandomState();
+        }
+
+        // Store the state, it may need to be accessed later.
+        $this->state = $options['state'];
+
+        $options += [
+            // Do not set the default state here! The random generator takes a
+            // non-trivial amount of time to run.
+            'response_type'   => 'code',
+            'approval_prompt' => 'auto',
+        ];
+
+        $scopes = is_array($this->scopes) ? implode($this->scopeSeparator, $this->scopes) : $this->scopes;
 
         $params = [
-            'client_id' => $this->clientId,
-            'redirect_uri' => $this->redirectUri,
-            'state' => $this->state,
-            'scope' => is_array($this->scopes) ? implode($this->scopeSeparator, $this->scopes) : $this->scopes,
-            'response_type' => isset($options['response_type']) ? $options['response_type'] : 'code',
-            'approval_prompt' => isset($options['approval_prompt']) ? $options['approval_prompt'] : 'auto',
+            'client_id'       => $this->clientId,
+            'redirect_uri'    => $this->redirectUri,
+            'state'           => $this->state,
+            'scope'           => $scopes,
+            'response_type'   => $options['response_type'],
+            'approval_prompt' => $options['approval_prompt'],
         ];
 
         return $this->urlAuthorize().'?'.$this->httpBuildQuery($params, '', '&');
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -90,7 +90,11 @@ abstract class AbstractProvider implements ProviderInterface
      */
     protected $httpBuildEncType = 1;
 
-    public function __construct($options = [], HttpAdapterInterface $httpClient = null)
+    /**
+     * @param array $options
+     * @param array $collaborators
+     */
+    public function __construct($options = [], array $collaborators = [])
     {
         foreach ($options as $option => $value) {
             if (property_exists($this, $option)) {
@@ -98,7 +102,10 @@ abstract class AbstractProvider implements ProviderInterface
             }
         }
 
-        $this->setHttpClient($httpClient ?: new CurlHttpAdapter());
+        if (empty($collaborators['httpClient'])) {
+            $collaborators['httpClient'] = new CurlHttpAdapter();
+        }
+        $this->setHttpClient($collaborators['httpClient']);
     }
 
     public function setHttpClient(HttpAdapterInterface $client)
--- a/test/src/Provider/AbstractProviderTest.php
+++ b/test/src/Provider/AbstractProviderTest.php
@@ -89,6 +89,14 @@ class AbstractProviderTest extends \PHPUnit_Framework_TestCase
         $this->assertSame($mockAdapter, $mockProvider->getHttpClient());
     }
 
+    public function testConstructorSetsRandomFactory()
+    {
+        $mockAdapter = m::mock('RandomLib\Factory');
+
+        $mockProvider = new MockProvider([], ['randomFactory' => $mockAdapter]);
+        $this->assertSame($mockAdapter, $mockProvider->getRandomFactory());
+    }
+
     public function testSetRedirectHandler()
     {
         $this->testFunction = false;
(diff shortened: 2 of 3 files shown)


src/Provider/AbstractProvider.php
```diff
@@ -98,7 +108,15 @@ public function __construct($options = [], HttpAdapterInterface $httpClient = nu
...
-        $this->setHttpClient($httpClient ?: new CurlHttpAdapter());
+        if (empty($collaborators['httpClient'])) {
+            $collaborators['httpClient'] = new CurlHttpAdapter();
+        }
+        $this->setHttpClient($collaborators['httpClient']);
+
+        if (empty($collaborators['randomFactory'])) {
+            $collaborators['randomFactory'] = new RandomFactory();
+        }
+        $this->setRandomFactory($collaborators['randomFactory']);
```
shadowhand: @ramsey using the `$collaborators` allows the end user to add their own sources and mixers to overload how states are generated. This should be more than sufficient to allow customization.

src/Provider/AbstractProvider.php
```diff
@@ -135,17 +176,46 @@ public function setScopes(array $scopes)
...
+        $this->state = $options['state'];
+
+        $options += [
+            // Do not set the default state here! The random generator takes a
+            // non-trivial amount of time to run.
+            'response_type'   => 'code',
+            'approval_prompt' => 'auto',
+        ];
+
+        $scopes = is_array($this->scopes) ? implode($this->scopeSeparator, $this->scopes) : $this->scopes;
```
shadowhand: Would it make sense to allow the `scopes` to be injected with `$options`? Seems like it could be a useful thing, but obviously out of scope with this PR.

test/src/Provider/AbstractProviderTest.php
```diff
@@ -185,6 +193,33 @@ public function getHeadersTest()
...
+
+        $this->assertArrayHasKey('state', $qs);
+        $this->assertSame($xstate, $qs['state']);
+
+        // Same test, but using the non-mock implementation
+        $url = $this->provider->getAuthorizationUrl();
+
+        parse_str(parse_url($url, PHP_URL_QUERY), $qs);
+
+        $this->assertRegExp('/[a-zA-Z0-9\/+]{32}\b/', $qs['state']);
```
shadowhand: Character range is the [RandomLib default](https://github.com/ircmaxell/RandomLib#generator-generatestringlength-characters--).

shadowhand:
cc @SammyK

src/Provider/AbstractProvider.php
ramsey: Rather than injecting the factory, I think we should allow them to inject the generator itself. So, they use the factory first to get a generator outside of this code, and then they pass that generator into this lib, as long as it follows the proper interface, as suggested by @ircmaxell in https://github.com/thephpleague/oauth2-client/issues/212#issuecomment-97058200

src/Provider/AbstractProvider.php
ramsey: You mean the `scopes` would be part of `$options` when passed into this method?

shadowhand: Yes.

src/Provider/AbstractProvider.php
shadowhand: The factory is far more flexible, and you can register any number of mixers and sources with it. By having the factory inside the class, providers could make use of it for their own needs, instead of being limited to a single generator.

test/src/Provider/AbstractProviderTest.php
SammyK: Is there a test that shows that the `+` and `/` characters will be properly URL-encoded once they are built into an authorization URL? And are there any concerns with decoding them on the callback URL?

src/Provider/AbstractProvider.php
```diff
@@ -98,7 +108,15 @@ public function __construct($options = [], HttpAdapterInterface $httpClient = nu
             }
         }
 
-        $this->setHttpClient($httpClient ?: new CurlHttpAdapter());
+        if (empty($collaborators['httpClient'])) {
```
SammyK: `empty()` should be `isset()`

src/Provider/AbstractProvider.php
```diff
@@ -98,7 +108,15 @@ public function __construct($options = [], HttpAdapterInterface $httpClient = nu
...
+            $collaborators['httpClient'] = new CurlHttpAdapter();
+        }
+        $this->setHttpClient($collaborators['httpClient']);
+
+        if (empty($collaborators['randomFactory'])) {
```
SammyK: Same here: should be `isset()`

src/Provider/AbstractProvider.php
```diff
@@ -135,17 +176,46 @@ public function setScopes(array $scopes)
...
+            ->getRandomFactory()
+            ->getMediumStrengthGenerator();
+
+        return $generator->generateString($length);
+    }
+
     public function getAuthorizationUrl(array $options = [])
     {
-        $this->state = isset($options['state']) ? $options['state'] : md5(uniqid(rand(), true));
+        if (empty($options['state'])) {
```
SammyK: Also should be `isset()` :)

src/Provider/AbstractProvider.php
shadowhand: Why?

test/src/Provider/AbstractProviderTest.php
shadowhand: Not necessary, because the query parameters are built using the `httpBuildQuery` method, which automatically encodes everything. No concerns about decode on the other side, at least nothing we can control. We could provide an explicit character set if that would be an improvement, but I'm not convinced it is.

src/Provider/AbstractProvider.php
shadowhand: `false` is not a valid `httpClient`.

src/Provider/AbstractProvider.php
shadowhand: `false` is not a valid `state`

ircmaxell: Why? `!isset()` is functionally identical to `empty()` in this usage...

src/Provider/AbstractProvider.php
ramsey: If we use the factory, then any library other than RandomLib will need to implement a factory that extends RandomLib's factory, or the user will need to do that on their own. My point is: if we only require that they pass in a Generator, then the user may simply wrap whatever other library they are using to adhere to the Generator interface and pass it in, without needing to also implement a factory.

src/Provider/AbstractProvider.php
ramsey: That sounds good.

src/Provider/AbstractProvider.php
ircmaxell: @shadowhand considering that any security sensitive usage should use a medium strength generator, there shouldn't be any need for other generators. In fact, I'd go as far as to say simply using a medium strength generator should be good enough for pretty much all usages...

src/Provider/AbstractProvider.php
SammyK: Neither is `foo`. :)

src/Provider/AbstractProvider.php
SammyK: I thought `empty()` with throw a notice error if an array key didn't exists and `isset()` wouldn't... but I just check it and it doesn't. Doh! Lol.

But the other reason I was going to propose is that the lib uses `isset` originally but doing a search in the code, the lib uses `isset` and `empty` quite interchangeably.

ircmaxell: `empty($foo)` is identical to `(!isset($foo) || !$foo)`. It does check for undefined variables as well...

ramsey: IIRC, there's some coding standard or best practices out there that prefers `isset()` over `empty()`, but I can't find that standard/practices right now, and it's not one that the League specifies.

test/src/Provider/AbstractProviderTest.php
SammyK: Cool - do you think we should add the test just in case `httpBuildQuery` gets toyed with in the future and doesn't properly escape the `state` param? Or is that being overly cautious? :)

src/Provider/AbstractProvider.php
shadowhand: The difference being that `false` would be detected as empty and the default adapter would be injected. I think it's really bad practice to use `!isset` instead of `empty`, unless you want to explicitly allow for `false` or `0` as a value. Way too often I see this kind of thing:

```php
if (isset($something) && $something !== false) { ... }
// can be replaced with a much more clear intent
if (!empty($something)) { ... }
```

test/src/Provider/AbstractProviderTest.php
ramsey: Probably overly cautious. If PHP breaks [http_build_query()](http://php.net/http_build_query), then we won't be the only ones affected. :-)

src/Provider/AbstractProvider.php
shadowhand: @ramsey we clearly have a misunderstanding. The RandomLib factory allows users to inject their own instances of Source/Mixer into the factory using [registerSource](https://github.com/ircmaxell/RandomLib/blob/338d2f594f7148983805ad56f47de7e47d84b2b9/lib/RandomLib/Factory.php#L157) and `registerMixer`. The generator relies on Source/Mixer to provide the actual values, the generator simply acts as a manager.

src/Provider/AbstractProvider.php
```diff
@@ -135,17 +176,46 @@ public function setScopes(array $scopes)
...
+     * Get a new random string to use for auth state.
+     *
+     * @param  integer $length
+     * @return string
+     */
+    protected function getRandomState($length = 32)
+    {
+        $generator = $this
+            ->getRandomFactory()
+            ->getMediumStrengthGenerator();
```
shadowhand: @ircmaxell we always use the medium generator. By injecting the factory, we allow the user to provide their own source/mixer.

src/Provider/AbstractProvider.php
ramsey: Yep. I had a misunderstanding of how RandomLib worked. Continue on. :-)

test/src/Provider/AbstractProviderTest.php
shadowhand: @SammyK part of the issue is that it can't be easily tested... the generated string is random, so there's no guarantee we'll get a string that contains `+` or `/`. And it's also not a proper unit, because it is testing a separate method `httpBuildQuery`.

src/Provider/AbstractProvider.php
SammyK: Good point. My point about the `foo` string is that there should probably be a check that it's a valid generator:

```php
if (!$collaborators['httpClient'] instanceof RandomFactory) {
    // throw new \InvalidArgumentException('Wha?');
}
```

shadowhand: @SammyK that's done by the type hint on the setter.

src/Provider/AbstractProvider.php
```diff
@@ -115,6 +133,29 @@ public function getHttpClient()
         return $client;
     }
 
+    /**
+     * Set the instance of the CSPRNG random generator factory to use.
+     *
+     * @param  RandomFactory $factory
+     * @return $this
+     */
+    public function setRandomFactory(RandomFactory $factory)
```
shadowhand: @SammyK this type hint will trigger an error if the parameter is not the correct type.

src/Provider/AbstractProvider.php
```diff
@@ -98,7 +108,15 @@ public function __construct($options = [], HttpAdapterInterface $httpClient = nu
...
+        }
+        $this->setHttpClient($collaborators['httpClient']);
+
+        if (empty($collaborators['randomFactory'])) {
+            $collaborators['randomFactory'] = new RandomFactory();
+        }
+        $this->setRandomFactory($collaborators['randomFactory']);
     }
 
     public function setHttpClient(HttpAdapterInterface $client)
```
shadowhand: @SammyK type hint ensures correct type.

src/Provider/AbstractProvider.php
SammyK: Oh yes! I missed that part. :)

src/Provider/AbstractProvider.php
shadowhand: @ramsey I'll do that as follow up PR, since it has nothing to do with the CSPRNG.

ramsey: Make sense.

shadowhand:
@ramsey I think there are no further concerns... this should be ready to go.

ramsey:
Thanks!

PR #279, opened 2015-05-02
Remove options dump from AccessToken exception

### LINKED ISSUE #253: Cleaner way to throw InvalidArgumentException
I think it's a bad idea to use print_r() directly to append the exception message. In my application, I have to use explode and extract the error message such as **This_authorization_code_has_expired_** rather than directly getting it.
https://github.com/thephpleague/oauth2-client/blob/6930258/src/Token/AccessToken.php#L38-L41

shadowhand:
Clean up the exception message.

Fixes #253.

### DIFF
--- a/src/Token/AccessToken.php
+++ b/src/Token/AccessToken.php
@@ -34,11 +34,11 @@ class AccessToken
      */
     public function __construct(array $options = null)
     {
-        if (! isset($options['access_token'])) {
-            throw new \InvalidArgumentException(
-                'Required option not passed: access_token'.PHP_EOL
-                .print_r($options, true)
-            );
+        if (empty($options['access_token'])) {
+            throw new InvalidArgumentException(sprintf(
+                'Required option not passed: "%s"',
+                'access_token'
+            ));
         }
 
         $this->accessToken = $options['access_token'];


src/Token/AccessToken.php
```diff
@@ -34,11 +34,11 @@ class AccessToken
...
     {
-        if (! isset($options['access_token'])) {
-            throw new \InvalidArgumentException(
-                'Required option not passed: access_token'.PHP_EOL
-                .print_r($options, true)
-            );
+        if (empty($options['access_token'])) {
+            throw new InvalidArgumentException(sprintf(
+                'Required option not passed: "%s"',
+                'access_token'
```
ramsey: Why do you need to use `sprintf()` here for a static value?

shadowhand: I don't _need_ to, but I prefer the way it reads and it makes it easier to copy the exception later. I can actually update this PR to use the new `RequiredParameterTrait` and then it will make more sense.

shadowhand: Actually I'm not sure I want to do that, as it will change the exception type from `InvalidArgument` to `BadMethodCall`. The other usage is for an actual method call, but this is for a constructor... what do you think, @ramsey ?

ramsey: I think `InvalidArgument` makes most sense here.

I understand the readability on the `sprintf()` call, but I think it's unnecessary to use a function call when it's not needed. We can leave it in for now, but we should eventually do some benchmarks and look at call graphs to see if we can identify any bottlenecks and clean them up (I'm not saying `sprintf()` is a bottleneck, but it's an extra function call that can be avoided).

PR #280, opened 2015-05-02
Create AbstractGrant for default grant types

shadowhand:
- documented `GrantInterface` and hint arguments
- add `AbstractGrant` to simplify grant classes
- use `AbstractGrant` for all default grants

### DIFF
--- a/src/Tool/RequiredParameterTrait.php
+++ b/src/Tool/RequiredParameterTrait.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace League\OAuth2\Client\Tool;
+
+use BadMethodCallException;
+
+trait RequiredParameterTrait
+{
+    /**
+     * Checks for a required parameter in a hash.
+     *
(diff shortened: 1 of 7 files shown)


src/Tool/RequiredParameterTrait.php
```diff
@@ -0,0 +1,41 @@
...
+{
+    /**
+     * Checks for a required parameter in a hash.
+     *
+     * @throws BadMethodCallException
+     * @param  string $name
+     * @param  array  $params
+     * @return void
+     */
+    private function checkRequiredParameter($name, array $params)
```
shadowhand: This same trait is useful for `AccessToken` and possibly other supporting classes, such as implementations of `UserInterface`.

shadowhand:
@ramsey build failure was because of a bad regex in the CSPRNG test. I've fixed that in #281. All other tests passed.

ramsey:
Looks good. Thanks!

PR #282, opened 2015-05-02
Remove extra user methods from abstract provider

shadowhand:
- documented `UserInterface`
- remove methods for user data fetching
- modify internals for creating `User` objects in providers
- update tests

Refs #267 and discussions on #274.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -407,11 +407,20 @@ abstract class AbstractProvider implements ProviderInterface
         }
     }
 
+    /**
+     * Generate a user object from a successful user details request.
+     *
+     * @param object $response
+     * @param AccessToken $token
+     * @return League\OAuth2\Client\Provider\UserInterface
+     */
+    abstract protected function prepareUserDetails(array $response, AccessToken $token);
+
     public function getUserDetails(AccessToken $token)
     {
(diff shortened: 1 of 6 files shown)


src/Provider/AbstractProvider.php
```diff
@@ -408,11 +407,20 @@ protected function setResultUid(array &$result)
...
     }
 
+    /**
+     * Generate a user object from a successful user details request.
+     *
+     * @param object $response
+     * @param AccessToken $token
+     * @return League\OAuth2\Client\Provider\UserInterface
+     */
+    abstract protected function prepareUserDetails(array $response, AccessToken $token);
```
shadowhand: Made this protected abstract, since it really has no purpose as a public method.

src/Provider/AbstractProvider.php
```diff
@@ -486,7 +480,7 @@ protected function fetchUserDetails(AccessToken $token)
     {
         $url = $this->urlUserDetails($token);
 
-        $request = $this->getAuthenticatedRequest(Request::METHOD_GET, $url, $token);
+        $request = $this->getAuthenticatedRequest(RequestInterface::METHOD_GET, $url, $token);
```
shadowhand: :fearful:

shadowhand:
@ramsey as per our discussion before. Better coverage of the request factory too.

PR #284, opened 2015-05-03
Create GrantFactory for default grant loading

shadowhand:
- add `GrantFactory` as a registry for grants
- move default grant generator to `GrantFactory`
- update `AbstractProvider` to use `GrantFactory`
- update tests

### DIFF
--- a/src/Grant/GrantFactory.php
+++ b/src/Grant/GrantFactory.php
@@ -1,0 +1,89 @@
+<?php
+
+namespace League\OAuth2\Client\Grant;
+
+class GrantFactory
+{
+    /**
+     * @var array
+     */
+    protected $registry = [];
+
+    /**
+     * Define a grant class in the registry.
+     *
+     * @param  string $name
+     * @param  string $class
+     * @return $this
+     */
+    public function setGrant($name, $class)
+    {
+        $this->registry[$name] = $this->checkGrant($class);
+
+        return $this;
+    }
+
+    /**
+     * Get a grant instance by name.
+     *
+     * If the grant has not be registered, a default grant will be loaded.
+     *
+     * @param  string $name
+     * @param  string $options
+     * @return GrantInterface
+     */
+    public function getGrant($name, array $options = [])
+    {
+        if (empty($this->registry[$name])) {
+            $this->registry[$name] = $this->getGrantClass($name);
+        }
+
+        $class = $this->registry[$name];
+
+        return new $class($options);
+    }
+
+    /**
+     * Guess a grant class from the name of the grant.
+     *
+     * @param  string $name
+     * @return string
+     */
+    protected function getGrantClass($name)
+    {
+        // PascalCase the grant. E.g: 'authorization_code' becomes 'AuthorizationCode'
+        $class = str_replace(' ', '', ucwords(str_replace(['-', '_'], ' ', $name)));
+        $class = 'League\\OAuth2\\Client\\Grant\\' . $class;
+
+        return $this->checkGrant($class);
+    }
+
+    /**
+     * Determine if a variable is a valid grant.
+     *
+     * @param  mixed $class
+     * @return boolean
+     */
+    public function isGrant($class)
+    {
+        return is_subclass_of($class, 'League\OAuth2\Client\Grant\GrantInterface');
+    }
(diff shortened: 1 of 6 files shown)


shadowhand:
Having the default generator outside of the abstract provider separates the concerns. The big win here is that it allows users to overload the default grant types by injecting a new `GrantFactory` **or** overloading them at runtime with `$provider->getGrantFactory()->setGrant(...)`.

src/Grant/GrantFactory.php
```diff
@@ -0,0 +1,89 @@
...
+     */
+    public function getGrant($name, array $options = [])
+    {
+        if (empty($this->registry[$name])) {
+            $this->registry[$name] = $this->getGrantClass($name);
+        }
+
+        $class = $this->registry[$name];
+
+        return new $class($options);
```
shadowhand: This new constructor signature allows for some future possibilities, though the current implementation in `AbstractProvider` doesn't pass anything in right now. I don't feel strongly about this change, and I'm not sure where the `$options` would come from inside of the provider.

In some ways, this feels like a bit of a poor man dependency injection... I wonder if it would be better for the registry to contain *instances* instead of *classes* so that the same grants are reused. Since they are stateless, it would make a lot more sense:

```php
$grant = new MyCustomGrant($options);
$provider->getGrantFactory()->setGrant('custom', $grant);
```

Having instances in the registry would mean that all grant types are singleton, but I don't see a problem with that.

src/Grant/GrantFactory.php
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /**
+     * Guess a grant class from the name of the grant.
+     *
+     * @param  string $name
+     * @return string
+     */
+    protected function getGrantClass($name)
```
shadowhand: If we used instances, this method would be modified to return a grant instance instead of a class.

src/Grant/GrantFactory.php
ramsey: I could go either way on this. As it is now, you're always getting a new object when you call `getGrant()`, even if you've called it before. So, I can see value in storing instances and treating them as singletons over creating new objects each time.

src/Grant/GrantFactory.php
ramsey: Makes sense.

ramsey: Would `setGrant()` also need to be modified in order to allow the user to instantiate and set a grant instance?

src/Grant/GrantFactory.php
```diff
@@ -0,0 +1,89 @@
...
+    {
+        // PascalCase the grant. E.g: 'authorization_code' becomes 'AuthorizationCode'
+        $class = str_replace(' ', '', ucwords(str_replace(['-', '_'], ' ', $name)));
+        $class = 'League\\OAuth2\\Client\\Grant\\' . $class;
```
ramsey: Does this mean the grant being used must always be in the `League\OAuth2\Client\Grant` namespace?

src/Grant/GrantFactory.php
shadowhand: Due to the way grants operate right now, there's no harm in having singletons. I just have a small worry that someone will create a custom grant type and make it stateful.

src/Grant/GrantFactory.php
shadowhand: Correct.

src/Grant/GrantFactory.php
shadowhand: Anything that is guessed by this method (the default grant names) has to be in the `League` namespace. But `setGrant` will take any FQCN.

src/Grant/GrantFactory.php
ramsey: Okay. We'll keep them singletons for now and see how people end up using them.

src/Grant/GrantFactory.php
ramsey: Okay. Perfect.

src/Grant/GrantFactory.php
shadowhand: Should I make a follow up PR to use instances instead of class names?

PR #289, opened 2015-05-07
Add Reddit provider

rtheunissen:
This pull request is to add a Reddit provider to the list of third party providers. I've used it, tested it, and so far everything seems to work as expected.


ramsey:
Thanks!

PR #291, opened 2015-05-08
Use the correct parameters for HttpClient::get and post

### LINKED ISSUE #287: [1.0] Request parameters not passed correctly to the HTTP client
I noticed something strange on the 1.0 branch: take a look at the arguments passed to the client in [AbstractProvider/getAccessToken](https://github.com/thephpleague/oauth2-client/blob/1.0/src/Provider/AbstractProvider.php#L275) and compare them to the function definition in [AbstractHttpAdapterTemplate](https://github.com/egeloen/ivory-http-adapter/blob/0.6.0/src/AbstractHttpAdapterTemplate.php#L50).

    $httpResponse = $client->post($this->urlAccessToken(), [
        'headers' => $this->getHeaders(),
        'body' => $requestParams,
    ]);

When I changed it locally to match the method header, everything worked as expected.

    $httpResponse = $client->post($this->urlAccessToken(), $this->getHeaders(), $requestParams);

shadowhand:
Fixes #287.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -283,18 +283,20 @@ abstract class AbstractProvider implements ProviderInterface
                 case 'GET':
                     // @codeCoverageIgnoreStart
                     // No providers included with this library use get but 3rd parties may
-                    $httpResponse = $client->get($this->urlAccessToken(), [
-                        'headers' => $this->getHeaders(),
-                        'query' => $requestParams,
-                    ]);
+                    $httpResponse = $client->get(
+                        $this->urlAccessToken(),
+                        $this->getHeaders(),
+                        $requestParams
+                    );
                     $response = (string) $httpResponse->getBody();
                     break;
                     // @codeCoverageIgnoreEnd
                 case 'POST':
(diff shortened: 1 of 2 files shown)


ramsey:
Are you sure this is the correct way to pass these parameters in Guzzle 5? The documentation shows the method signature as that which you are replacing:

http://docs.guzzlephp.org/en/latest/http-messages.html#requests

ramsey:
Doh! Nevermind. This is due to the Ivory adapter, and clearly I was only looking at the Guzzle documentation. :facepalm:

rtheunissen:
Yup. https://github.com/guzzle/guzzle/blob/master/src/Client.php#L123

rtheunissen:
Yup we were passing it as if we were passing it directly to Guzzle, rather than the ivory adapter. I think?

shadowhand:
Exactly, we forgot to modify the `get` and `post` calls to match ivory syntax.

shadowhand:
@ramsey updated.

ramsey:
Thanks!

PR #292, opened 2015-05-08
Ensure that grant_type is the expected value

### LINKED ISSUE #288: [1.0] Reddit complained that it received an invalid grant type
On 1.0, I was receiving 401's from Reddit stating that I was using in invalid grant type. I'm not sure where in the pipeline it was failing, but the *object* grant in [AbstractProvider/getAccessToken]() doesn't seem to ever be requested its __toString(). As soon as I changed $grant to "$grant" the problem disappeared.

Tried to have a look through egeloen/ivory-http-adapter but didn't spend too much time trying to figure out where the translation actually happens.

shadowhand:
Force `Grant::__toString` to be executed when using the grant as a request parameter.

Fixes #288

### DIFF
--- a/test/src/Provider/AbstractProviderTest.php
+++ b/test/src/Provider/AbstractProviderTest.php
@@ -250,6 +250,51 @@ class AbstractProviderTest extends \PHPUnit_Framework_TestCase
         $this->assertRegExp('/^[a-zA-Z0-9\/+]{32}$/', $qs['state']);
     }
 
+    public function testGetAccessToken()
+    {
+        $provider = new MockProvider([
+          'clientId' => 'mock_client_id',
+          'clientSecret' => 'mock_secret',
+          'redirectUri' => 'none',
+        ]);
+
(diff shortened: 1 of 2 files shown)


test/src/Provider/AbstractProviderTest.php
```diff
@@ -250,6 +250,51 @@ public function testRandomGeneratorCreatesRandomState()
         $this->assertRegExp('/^[a-zA-Z0-9\/+]{32}$/', $qs['state']);
     }
 
+    public function testGetAccessToken()
```
shadowhand: Oddly enough, we didn't have a test that verified a successful token fetch.

shadowhand:
@ramsey let's merge this before #291 so that I can update that PR to add the additional `with` validation on the request.

ramsey:
I'm on the road, so I'll go ahead and merge this in so that I'm not holding you back from updating #291. I'll merge in #291 tonight.

PR #294, opened 2015-05-09
Refactor internal configuration and public properties

shadowhand:
- all properties in `AbstractProvider` made protected
- many config properties removed in favor of protected methods that can be overloaded
- modified header, scope, redirects, etc handling
- tests updated

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -16,12 +16,17 @@ use UnexpectedValueException;
 abstract class AbstractProvider implements ProviderInterface
 {
     /**
+     * @var string HTTP method used to fetch access tokens.
+     */
+    const ACCESS_TOKEN_METHOD = 'post';
+
+    /**
      * @var string Separator used for authorization scopes.
      */
     const SCOPE_SEPARATOR = ',';
 
     /**
-     * @var string Separator used for authorization scopes.
+     * @var string
      */
     protected $clientId = '';
 
@@ -49,11 +54,6 @@ abstract class AbstractProvider implements ProviderInterface
      * @var string
      */
     protected $uidKey = 'uid';
-
-    /**
-     * @var string
-     */
-    protected $method = 'post';
 
     /**
      * @var string
@@ -294,7 +294,7 @@ abstract class AbstractProvider implements ProviderInterface
 
         try {
             $client = $this->getHttpClient();
-            switch (strtoupper($this->method)) {
+            switch (strtoupper(static::ACCESS_TOKEN_METHOD)) {
                 case 'GET':
                     // @codeCoverageIgnoreStart
                     // No providers included with this library use get but 3rd parties may
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -18,62 +18,62 @@ abstract class AbstractProvider implements ProviderInterface
     /**
      * @var string
      */
-    public $clientId = '';
+    protected $clientId = '';
 
     /**
      * @var string
      */
-    public $clientSecret = '';
+    protected $clientSecret = '';
 
     /**
      * @var string
      */
-    public $redirectUri = '';
+    protected $redirectUri = '';
 
     /**
      * @var string
      */
-    public $state;
+    protected $state;
 
     /**
      * @var string
      */
-    public $name;
+    protected $name;
 
     /**
      * @var string
      */
-    public $uidKey = 'uid';
+    protected $uidKey = 'uid';
 
     /**
      * @var array
      */
-    public $scopes = [];
+    protected $scopes = [];
 
     /**
      * @var string
      */
-    public $method = 'post';
+    protected $method = 'post';
 
     /**
      * @var string
      */
-    public $scopeSeparator = ',';
+    protected $scopeSeparator = ',';
 
     /**
      * @var string
      */
-    public $responseType = 'json';
+    protected $responseType = 'json';
 
     /**
      * @var array
      */
-    public $headers = [];
+    protected $headers = [];
 
     /**
      * @var string
      */
-    public $authorizationHeader;
+    protected $authorizationHeader;
 
     /**
      * @var GrantFactory
@@ -178,6 +178,18 @@ abstract class AbstractProvider implements ProviderInterface
     public function getRandomFactory()
     {
         return $this->randomFactory;
+    }
+
+    /**
+     * Get the current state of the OAuth flow.
+     *
+     * This can be accessed by the redirect handler during authorization.
+     *
+     * @return string
+     */
+    public function getState()
+    {
+        return $this->state;
     }
 
     // Implementing these interfaces methods should not be required, but not
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -21,6 +21,11 @@ abstract class AbstractProvider implements ProviderInterface
     const ACCESS_TOKEN_METHOD = 'post';
 
     /**
+     * @var string Type of response expected from the provider.
+     */
+    const RESPONSE_TYPE = 'json';
+
+    /**
      * @var string Separator used for authorization scopes.
      */
     const SCOPE_SEPARATOR = ',';
@@ -54,11 +59,6 @@ abstract class AbstractProvider implements ProviderInterface
      * @var string
      */
     protected $uidKey = 'uid';
-
-    /**
-     * @var string
-     */
-    protected $responseType = 'json';
 
     /**
      * @var GrantFactory
@@ -387,7 +387,7 @@ abstract class AbstractProvider implements ProviderInterface
     {
         $result = [];
 
-        switch ($this->responseType) {
+        switch (static::RESPONSE_TYPE) {
             case 'json':
                 $result = json_decode($response, true);
                 if (JSON_ERROR_NONE !== json_last_error()) {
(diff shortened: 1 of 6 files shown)


src/Provider/AbstractProvider.php
```diff
@@ -213,27 +188,47 @@ protected function getRandomState($length = 32)
...
+    abstract protected function getDefaultScopes();
+
+    /**
+     * Get the scope separator used by this provider.
+     *
+     * @return string
+     */
+    protected function getScopeSeparator()
+    {
+        return ',';
```
shadowhand: I thought about making this a class constant.

src/Provider/AbstractProvider.php
```diff
@@ -260,6 +253,18 @@ public function authorize(array $options = [])
...
+    /**
+     * Get the request method used for fetching request tokens.
+     *
+     * Supported defaults are either "post" or "get".
+     *
+     * @return string
+     */
+    protected function getAccessTokenMethod()
+    {
+        return 'post';
```
shadowhand: This too could be a constant.

src/Provider/AbstractProvider.php
```diff
@@ -362,6 +368,18 @@ public function getResponse(RequestInterface $request)
...
     /**
+     * Get the type of the response for this provider.
+     *
+     * Supported defaults are either "json" or "string".
+     *
+     * @return string
+     */
+    protected function getResponseType()
+    {
+        return 'json';
```
shadowhand: Also could be a constant.

src/Provider/AbstractProvider.php
```diff
@@ -399,34 +417,28 @@ protected function parseResponse($response)
...
-     * @param  array $result
-     * @return array
+     * @return string
      */
-    protected function prepareAccessTokenResult(array $result)
+    protected function getUidKey()
     {
-        $this->setResultUid($result);
-        return $result;
+        return 'uid';
```
shadowhand: Another possibility for a constant.

src/Tool/BearerAuthorizationTrait.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace League\OAuth2\Client\Tool;
+
+/**
+ * Enables `Bearer` header authorization for providers.
+ *
+ * http://tools.ietf.org/html/rfc6750
+ */
+trait BearerAuthorizationTrait
```
shadowhand: Usage of this is in the `Fake` provider used for testing.

src/Tool/MacAuthorizationTrait.php
```diff
@@ -0,0 +1,55 @@
...
+    abstract protected function getMacSignature($id, $ts, $nonce);
+
+    // AbstractProvider
+    abstract protected function getRandomState($length);
+
+    protected function getAuthorizationHeaders($token = null)
+    {
+        // This is currently untested and provided only as an example. If you
+        // complete the implementation, please create a pull request for
+        // https://github.com/thephpleague/oauth2-client
```
shadowhand: Thought it would be worthwhile to include this, even though it is only a stub. Hopefully someone will pick it up.

test/src/Provider/AbstractProviderTest.php
```diff
@@ -215,6 +205,13 @@ public function getHeadersTest()
 
     public function testScopesOverloadedDuringAuthorize()
     {
+        $url = $this->provider->getAuthorizationUrl();
+
+        parse_str(parse_url($url, PHP_URL_QUERY), $qs);
+
+        $this->assertArrayHasKey('scope', $qs);
+        $this->assertSame('test', $qs['scope']);
```
shadowhand: Adds test for default scope.

test/src/Provider/AbstractProviderTest.php
```diff
@@ -299,6 +296,9 @@ public function testGetAccessToken()
         $result = $provider->getAccessToken($grant, ['code' => 'mock_authorization_code']);
 
         $this->assertSame($result, $token);
+        $this->assertSame($raw_response['uid'], $token->uid);
+        $this->assertSame($raw_response['access_token'], $token->accessToken);
+        $this->assertSame($raw_response['expires'], $token->expires);
```
shadowhand: Maybe these tests shouldn't be included? I added them because I wanted verification that the UID changes were done right.

test/src/Provider/AbstractProviderTest.php
```diff
@@ -403,8 +403,7 @@ public function testAuthenticatedRequestAndResponse()
     {
         $token = new AccessToken(['access_token' => 'abc', 'expires_in' => 3600]);
 
-        $provider = clone $this->provider;
-        $provider->authorizationHeader = 'Bearer';
+        $provider = new MockProvider(['authorizationHeader' => 'Bearer']);
```
shadowhand: Oops I need to remove this.

test/src/Provider/Fake.php
```diff
@@ -3,11 +3,14 @@
 namespace League\OAuth2\Client\Test\Provider;
 
 use League\OAuth2\Client\Token\AccessToken;
+use League\OAuth2\Client\Tool\BearerAuthorizationTrait;
 use League\OAuth2\Client\Provider\AbstractProvider;
 use League\OAuth2\Client\Provider\Exception\IdentityProviderException;
 
 class Fake extends AbstractProvider
 {
+    use BearerAuthorizationTrait;
```
shadowhand: Usage of the new traits for auth headers.

src/Provider/AbstractProvider.php
```diff
@@ -18,62 +18,22 @@
...
 
     /**
      * @var string
      */
-    public $state;
-
-    /**
-     * @var string
-     */
-    public $name;
```
shadowhand: I didn't see this being used anywhere... is there value in keeping it/

src/Provider/AbstractProvider.php
```diff
@@ -16,64 +16,44 @@
...
      */
-    public $redirectUri = '';
+    const RESPONSE_TYPE = 'json';
 
     /**
-     * @var string
+     * @var string Separator used for authorization scopes.
      */
-    public $state;
+    const SCOPE_SEPARATOR = ',';
```
shadowhand: Converted the constant value to... constants.

src/Provider/AbstractProvider.php
```diff
@@ -399,37 +402,21 @@ protected function parseResponse($response)
...
+        $result['uid'] = $result[static::ACCESS_TOKEN_UID];
         return $result;
     }
 
     /**
-     * Sets any result keys we've received matching our provider-defined uidKey to the key "uid".
-     *
-     * @param array $result
-     */
-    protected function setResultUid(array &$result)
```
shadowhand: I didn't see a reason to have this entirely separate method when we have `prepareAccessTokenResult` instead. I think it was vestigial from methods added <1.0.

src/Provider/AbstractProvider.php
```diff
@@ -479,26 +466,49 @@ protected function fetchUserDetails(AccessToken $token)
...
+     * http://tools.ietf.org/html/rfc6749#section-7.1
+     *
+     * No default is provided, providers must overload this method to activate
+     * authorization headers.
+     *
+     * @return array
+     */
+    protected function getAuthorizationHeaders($token = null)
+    {
+        return [];
```
shadowhand: I think this is much better than the previous way it was done, because it allows using custom headers, not just `Authorize`. Some APIs use `X-Token` for example.

rtheunissen:
This looks good to me. As long as we keep `getHeaders` public as well I'm all for it.

ramsey:
All looks great to me. Looks like you said you needed to remove/change some things. Is it all good to merge, or should I wait?

shadowhand:
Okay to merge. I made my changes immediately after opening the PR.

rtheunissen:
What are your thoughts on making provider configuration immutable?

rtheunissen:
Actually nevermind I just had a closer look. :)

src/Provider/AbstractProvider.php
```diff
@@ -16,64 +16,44 @@
...
      */
-    public $clientSecret = '';
+    const ACCESS_TOKEN_UID = 'uid';
 
     /**
-     * @var string
+     * @var string Type of response expected from the provider.
      */
-    public $redirectUri = '';
+    const RESPONSE_TYPE = 'json';
```
SammyK: Having the response type as a constant is problematic for the Facebook provider. In Graph v2.2 and below the response type was "string" and starting in v2.3 the response type was "json". See how the [Facebook provider was dealing with this issue](https://github.com/thephpleague/oauth2-facebook/blob/master/src/Provider/Facebook.php#L130-L144). Any way we can revert this change for the response type?

Also - although I like constants, there's no guarantee that any of these constants will be consistent across all versions of a provider's API (as in the response type example). Perhaps we should revert all of them to be on the safe side? :)

philsturgeon:
Make it method and shove it in the interface then?

rtheunissen:
> Perhaps we should revert all of them to be on the safe side? :)
So have protected methods return the default, and be overridden where necessary? That's how I've always done things as it has all the benefits of the property not being accessible outside of the class (and its descendants), having a default value present, and being flexible for extending classes. You can then return a response type based on the value of something else, eg. Graph API version.

No need to have these in the interface as extending classes would only override if required to do so.

src/Provider/AbstractProvider.php
shadowhand: thephpleague/oauth2-client#301

PR #298, opened 2015-05-12
Make all properties of AccessToken protected

shadowhand:
- add getter methods
- update tests

Refs #294


ramsey:
Good stuff! I like it.

shadowhand:
:clap:

PR #299, opened 2015-05-13
Fixed improper use of the word "encryption"

jeteon:
I think what is meant here is encoding rather than encryption.


ramsey:
Thanks!

PR #301, opened 2015-05-14
Allow response type to be determined at runtime

shadowhand:
Partial revert of #294 to address providers (Facebook) that changed response type between API versions.

Refs thephpleague/oauth2-facebook#2.

cc @SammyK


SammyK:
:+1: Love it. :)

ramsey:
Looks great. Thanks!

PR #305, opened 2015-05-16
Add "timeout" option for providers

### LINKED ISSUE #303: [1.0] Add option to overwrite HTTP client timeout
The default timeout for the curl adapter is 10 seconds which isn't enough time for low Starbucks connections. :)

I don't have time to add this myself, but it should be low-hanging fruit to anyone who wants to take it on. The constructor of the `AbstractProvider` needs something like this:

```php
        if (empty($collaborators['httpClient'])) {
            $client = new CurlHttpAdapter();
            $config = $client->getConfiguration();
            $config->setTimeOut(60); // This should be configurable with an option
            $collaborators['httpClient'] = $client;
        }
```

Then we can have a `timeout` option to overwrite the default when we're on slower connections:

```php
$provider = FooProvider([
    // . . . 
    'timeout' => 60,
]);
```

This is really needed when using the authenticated client to upload videos. :)

shadowhand:
Passed through to HTTP client configuration.

Fixes #303

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -107,6 +107,12 @@ abstract class AbstractProvider implements ProviderInterface
             $collaborators['randomFactory'] = new RandomFactory();
         }
         $this->setRandomFactory($collaborators['randomFactory']);
+
+        if (!empty($options['timeout'])) {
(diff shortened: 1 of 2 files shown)


SammyK:
:+1:

PR #306, opened 2015-05-16
Ignore exceptions that have no response attached

### LINKED ISSUE #304: [1.0] Improve error handling when HTTP client fails
When testing out the Facebook provider on a slow connection, I kept getting this error after 10 seconds:

> Fatal error: Call to a member function getBody() on null in AbstractProvider.php on line 316

The issue was due to a low timeout (10 seconds) that caused the HTTP client to barf out a `HttpAdapterException` on [these lines](https://github.com/thephpleague/oauth2-client/blob/1.0/src/Provider/AbstractProvider.php#L315-L317). And [these lines are also affected](https://github.com/thephpleague/oauth2-client/blob/1.0/src/Provider/AbstractProvider.php#L363-L365) by this bug.

1. We need better error handling of when the provider returns an error response vs when the HTTP client dies due to a timeout or something.
2. The methods that use the HTTP client should throw an exception if the HTTP client dies.

I don't have time to add this unfortunately but it should be a relatively easy PR. :)

shadowhand:
When HTTP client fails and no response is available, allow the exception to bubble.

Fixes #304

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -313,6 +313,9 @@ abstract class AbstractProvider implements ProviderInterface
                 // @codeCoverageIgnoreEnd
             }
         } catch (HttpAdapterException $e) {
+            if (!$e->hasResponse()) {
+                throw $e;
+            }
             $response = (string) $e->getResponse()->getBody();
         }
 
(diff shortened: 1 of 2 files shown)


src/Provider/AbstractProvider.php
```diff
@@ -361,6 +364,9 @@ public function getResponse(RequestInterface $request)
 
             $response = (string) $httpResponse->getBody();
         } catch (HttpAdapterException $e) {
+            if (!$e->hasResponse()) {
+                throw $e;
```
shadowhand: @SammyK do you think this is alright, or should we wrap the exception with a League version? It seems unnecessary to me, since this represents something the provider can't recover from.

SammyK: :+1: I'm a big fan of letting exceptions like this bubble up. :)

ramsey:
I'm a fan, too. :+1:

PR #315, opened 2015-06-01
Switch to Guzzle 6 for HTTP client

### LINKED ISSUE #314: Switch to Guzzle 6
@ramsey do we want to switch to [Guzzle 6](http://docs.guzzlephp.org/en/latest/psr7.html)? It looks like the Ivory adapter is still using an outdated PSR-7 interface. I've experimented with integrating Guzzle 6 and the code is actually smaller than using Ivory and the behavior feels more consistent.

shadowhand:
Replaces Ivory HTTP adapter with Guzzle 6. Moves towards better PSR-7 compliance.

Also bumps the PHP version requirement to 5.5, because Guzzle 6 requires it.

Fixes #314

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,8 +5,8 @@
     "require": {
         "php": ">=5.4.0",
         "ext-curl": "*",
-        "egeloen/http-adapter": "0.6.*",
-        "ircmaxell/random-lib": "~1.1"
+        "ircmaxell/random-lib": "~1.1",
+        "guzzlehttp/guzzle": "~6.0"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.0",
--- a/composer.json
+++ b/composer.json
@@ -3,7 +3,7 @@
     "description": "OAuth 2.0 Client Library",
     "license": "MIT",
     "require": {
-        "php": ">=5.4.0",
+        "php": ">=5.5.0",
         "ext-curl": "*",
         "ircmaxell/random-lib": "~1.1",
         "guzzlehttp/guzzle": "~6.0"
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -3,13 +3,14 @@
 namespace League\OAuth2\Client\Provider;
 
 use Closure;
-use Ivory\HttpAdapter\CurlHttpAdapter;
-use Ivory\HttpAdapter\HttpAdapterException;
-use Ivory\HttpAdapter\HttpAdapterInterface;
-use Ivory\HttpAdapter\Message\RequestInterface;
+use GuzzleHttp\Client as HttpClient;
+use GuzzleHttp\ClientInterface as HttpClientInterface;
+use GuzzleHttp\Exception\BadResponseException;
+use Psr\Http\Message\RequestInterface;
 use League\OAuth2\Client\Provider\Exception\IdentityProviderException;
 use League\OAuth2\Client\Grant\GrantFactory;
 use League\OAuth2\Client\Token\AccessToken;
+use League\OAuth2\Client\Tool\RequestFactory;
 use RandomLib\Factory as RandomFactory;
 use UnexpectedValueException;
 
@@ -28,7 +29,7 @@ abstract class AbstractProvider implements ProviderInterface
     /**
      * @var string HTTP method used to fetch access tokens.
      */
-    const ACCESS_TOKEN_METHOD = 'post';
+    const ACCESS_TOKEN_METHOD = 'POST';
 
     /**
      * @var string Key used in the access token response to identify the user.
@@ -66,6 +67,11 @@ abstract class AbstractProvider implements ProviderInterface
     protected $grantFactory;
 
     /**
+     * @var RequestFactory
+     */
+    protected $requestFactory;
+
+    /**
      * @var HttpAdapterInterface
      */
     protected $httpClient;
@@ -98,8 +104,16 @@ abstract class AbstractProvider implements ProviderInterface
         }
         $this->setGrantFactory($collaborators['grantFactory']);
 
+        if (empty($collaborators['requestFactory'])) {
+            $collaborators['requestFactory'] = new requestFactory();
+        }
+        $this->setRequestFactory($collaborators['requestFactory']);
+
         if (empty($collaborators['httpClient'])) {
-            $collaborators['httpClient'] = new CurlHttpAdapter();
+            $client_options = ['timeout'];
+            $collaborators['httpClient'] = new HttpClient(
+                array_intersect_key($options, array_flip($client_options))
+            );
         }
         $this->setHttpClient($collaborators['httpClient']);
 
@@ -107,12 +121,6 @@ abstract class AbstractProvider implements ProviderInterface
             $collaborators['randomFactory'] = new RandomFactory();
         }
         $this->setRandomFactory($collaborators['randomFactory']);
-
-        if (!empty($options['timeout'])) {
-            $this->getHttpClient()
-                ->getConfiguration()
-                ->setTimeout($options['timeout']);
-        }
     }
 
     /**
@@ -139,12 +147,35 @@ abstract class AbstractProvider implements ProviderInterface
     }
 
     /**
-     * Set the HTTP adapter instance.
+     * Set the request factory instance.
      *
-     * @param  HttpAdapterInterface $client
+     * @param  RequestFactory $factory
      * @return $this
      */
-    public function setHttpClient(HttpAdapterInterface $client)
+    public function setRequestFactory(RequestFactory $factory)
+    {
+        $this->requestFactory = $factory;
+
+        return $this;
+    }
+
+    /**
+     * Get the request factory instance.
+     *
+     * @return RequestFactory
+     */
+    public function getRequestFactory()
+    {
+        return $this->requestFactory;
+    }
+
+    /**
+     * Set the HTTP adapter instance.
+     *
+     * @param  HttpClientInterface $client
+     * @return $this
+     */
+    public function setHttpClient(HttpClientInterface $client)
     {
         $this->httpClient = $client;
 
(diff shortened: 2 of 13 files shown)


shadowhand:
Depends on guzzle/guzzle#1100 being resolved.

src/Provider/AbstractProvider.php
```diff
@@ -291,51 +288,63 @@ public function getAccessToken($grant = 'authorization_code', array $params = []
...
+     * Available options are described here: http://docs.guzzlephp.org/en/latest/request-options.html
+     *
+     * @param  string $method
+     * @param  string $url
+     * @parma  array  $options
+     * @return GuzzleHttp\Psr7\Request
+     */
+    protected function getGuzzleRequest($method, $url, array $options = [])
+    {
+        return $this->getHttpClient()->requestFactory($method, $url, $options);
```
shadowhand: @mtdowling here we use theoretical Guzzle request factory. Note that the interface is not public, the fact that this is Guzzle doesn't matter.

src/Provider/AbstractProvider.php
```diff
@@ -291,51 +288,63 @@ public function getAccessToken($grant = 'authorization_code', array $params = []
...
+     * Creates a PSR-7 compatible request instance that can be modified.
+     * The request is not automatically authenticated.
+     *
+     * @param  string $method
+     * @param  string $url
+     * @return RequestInterface
+     */
+    public function getRequest($method, $url)
+    {
+        return $this->getGuzzleRequest($method, $url);
```
shadowhand: And this is the public interface that gets a generic PSR-7 request instance. It just happens to be backed by Guzzle at the moment.

src/Provider/AbstractProvider.php
```diff
@@ -344,13 +353,8 @@ public function getAccessToken($grant = 'authorization_code', array $params = []
...
     {
-        $factory = $this->getHttpClient()
-            ->getConfiguration()
-            ->getMessageFactory();
-
-        $request = $factory->createRequest($url, $method);
-        $request->addHeaders($this->getHeaders($token));
-        return $request;
+        $options = ['headers' => $this->getHeaders($token)];
+        return $this->getGuzzleRequest($method, $url, $options);
```
shadowhand: And here we have a very convenient method that gives us back an authenticated request, also configured (privately) by Guzzle.

src/Provider/AbstractProvider.php
```diff
@@ -364,21 +368,14 @@ public function getAuthenticatedRequest($method, $url, AccessToken $token)
...
-
-            $httpResponse = $client->sendRequest($request);
-
-            $response = (string) $httpResponse->getBody();
-        } catch (HttpAdapterException $e) {
-            if (!$e->hasResponse()) {
-                throw $e;
-            }
-            $response = (string) $e->getResponse()->getBody();
+            $response = $this->getHttpClient()->send($request);
```
shadowhand: And here we have a very convenient method to send a request and get back a response. It's currently backed by Guzzle, but that is completely hidden from the end user.

rtheunissen:
Guzzle 6 requires PHP >= 5.5, so we would need to bump the minimum version in composer.json otherwise Composer will fail to install for someone running 5.4.

shadowhand:
Some of the tests are failing but this is closer to functional.

src/Provider/AbstractProvider.php
```diff
@@ -289,68 +320,63 @@ public function getAccessToken($grant = 'authorization_code', array $params = []
...
-                    );
-                    $response = (string) $httpResponse->getBody();
-                    break;
+        switch (strtoupper(static::ACCESS_TOKEN_METHOD)) {
+            case 'GET':
                 // @codeCoverageIgnoreStart
-                default:
-                    throw new \InvalidArgumentException('Neither GET nor POST is specified for request');
+                // No providers included with this library use get but 3rd parties may
+                $urlAccessToken .= (strpos($urlAccessToken, '?') ? '&' : '?') . $requestParams;
```
shadowhand: I really don't like this. Hopefully we can find a better way.

shadowhand:
@rtheunissen @ramsey this is now complete and tested. Code coverage has increased slightly.

src/Tool/RequestFactory.php
```diff
@@ -0,0 +1,55 @@
...
+     * Get a request using a simplified array of options.
+     *
+     * @param  null|string $method
+     * @param  null|string $uri
+     * @param  array $options
+     * @return Request
+     */
+    public function getRequestWithOptions($method, $uri, array $options = [])
+    {
+        $func    = new \ReflectionMethod($this, 'getRequest');
```
shadowhand: Are we okay with using Reflection in this method? It seemed more DRY than the alternative, but maybe it breaks HHVM?

rtheunissen: [Using array merge instead, which is not as DRY but doesn't use reflection.](https://github.com/shadowhand/oauth2-client/pull/3/files#r32169493)

src/Provider/AbstractProvider.php
rtheunissen: [Using parse_url to determine if a query is present](https://github.com/shadowhand/oauth2-client/pull/3/files#r32168988)

composer.json
```diff
@@ -3,10 +3,10 @@
     "description": "OAuth 2.0 Client Library",
     "license": "MIT",
     "require": {
-        "php": ">=5.4.0",
+        "php": ">=5.5.0",
```
rtheunissen: Nice :+1:

src/Provider/AbstractProvider.php
shadowhand: Are you saying that a string search is less efficient than parsing a query string?

rtheunissen: > parsing a query string

You're parsing the URL, which of course wouldn't be more efficient than a string search (micro-optimization anyway). It's just more reliable I would think, than to assume that the presence of a '?' guarantees we should use &. What if I use 'http://test.com/?'? We don't want 'http://test.com/?&param=value.

shadowhand: Seems like an odd corner case, but okay.

rtheunissen: I'd rather cater for the odd corner case than get a micro-level speed increase, especially on a method that will most likely just be called once for every request (not even that, seeing as a token would most likely be stored somewhere).

ramsey:
Hmm... the failure message on HHVM is:

> Failed asserting that exception of type "BadMethodCallException" matches expected exception "\League\OAuth2\Client\Provider\Exception\IdentityProviderException". Message was: "Call to a member function mockery_getName() on a non-object (NULL)"

It's the `testClientErrorTriggersProviderException()` test where this error is occurring. So, I don't think it's the `RelfectionMethod` that's causing the issue. That said, I'm wary of using reflection in code intended to run in production. It tends to slow things down. Can we work around it?

rtheunissen:
@ramsey see https://github.com/shadowhand/oauth2-client/pull/3/files#r32169493

ramsey:
After reviewing the reflection code in more depth, I agree with @rtheunissen about switching to use `array_merge()` here. I understand that you're trying to avoid defining the parameters and their default values twice, but the use of reflection is pretty heavy-handed.

shadowhand:
> the use of reflection is pretty heavy-handed

Fair enough. I didn't think it was all the brilliant anyways.

shadowhand:
@ramsey assuming the tests pass, this is ready to go.

ramsey:
@shadowhand Build failing again in HHVM. Same issue. Odd.

rtheunissen:
Couldn't find anything useful on Google either. At least it's only specific to that test, so shouldn't be too hard to figure out what's causing it.

rtheunissen:
@shadowhand See if merging [this commit](https://github.com/rtheunissen/oauth2-client/commit/896c2ea39611f9679532ee297f95cc7907e8e4f1) fixes it?

rtheunissen:
@shadowhand @ramsey I have fixed the hhvm issue. I'm not 100% sure why it was occurring but if we adjust the way we create the expected exception, it works. See [this commit](https://github.com/rtheunissen/oauth2-client/commit/48941502284d50650c4838c48f8d4f9f18f392b0).

shadowhand:
@rtheunissen thanks! Merged it.

rtheunissen:
Thanks :+1:

ramsey:
I'll try to dig into this tonight and see if I can figure out what's going on in the test with HHVM.

ramsey:
The test result I'm seeing in Travis appears to be before @shadowhand merged in @rtheunissen's fix. @shadowhand, have you pushed your latest changes to your branch?

shadowhand:
@ramsey I did push it, but it broke on something else. I think it will be fixed now... we'll see.

shadowhand:
@ramsey HHVM worked, 5.5 stalled for some reason https://travis-ci.org/thephpleague/oauth2-client/jobs/66405399 I think we're good to go. You could restart the build if we want to double check 5.5.

stevenmaguire:
Having just reviewed the docs, and since this change will bump the minimum php support to 5.5, the README should be updated to remove acknowledgement for support of 5.4

shadowhand:
@stevenmaguire thanks for noticing that, I've amended that commit.

rtheunissen:
@shadowhand I think you're missing [this commit](https://github.com/rtheunissen/oauth2-client/commit/48941502284d50650c4838c48f8d4f9f18f392b0)? It's the one that actually fixed it for me running hhvm locally.

shadowhand:
@rtheunissen no the build is fine, 5.5 is timing out on Travis, but I don't think it is because of anything in this PR.

ramsey:
I ran the 5.5 build again, and it's passing now. Yay!

ramsey:
Going to go ahead and merge this in. Let's start getting things cleaned up (if needed), and ready for a release. Wooo!

rtheunissen:
@shadowhand Okay good to know, sorry about the confusion. :<

@ramsey Thanks for your help, keep in mind there are still a few steps to take before release. I have maybe 4 PR's ready, and there's also the gh-pages branch which isn't crucial but I think would be nice for 1.0.

ramsey:
No worries. I'm not rushing a release. I'm just saying we're very close. :-)

PR #317, opened 2015-06-02
Providing a PSR-7 Request instance

rtheunissen:
This is primarily a proof of concept to complement #315, which provides a way to create an authorized request, which the client or something else can send.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -3,10 +3,12 @@
 namespace League\OAuth2\Client\Provider;
 
 use Closure;
-use Ivory\HttpAdapter\CurlHttpAdapter;
-use Ivory\HttpAdapter\HttpAdapterException;
-use Ivory\HttpAdapter\HttpAdapterInterface;
-use Ivory\HttpAdapter\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\RequestInterface;
+use GuzzleHttp\Psr7\Request;
+use GuzzleHttp\ClientInterface;
+use GuzzleHttp\Client;
+use GuzzleHttp\Exception\RequestException;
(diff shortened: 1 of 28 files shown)


src/Provider/AbstractProvider.php
```diff
@@ -260,31 +258,15 @@ public function getAccessToken($grant = 'authorization_code', array $params = []
...
-                // @codeCoverageIgnoreStart
-                default:
-                    throw new \InvalidArgumentException('Neither GET nor POST is specified for request');
-                // @codeCoverageIgnoreEnd
-            }
-        } catch (HttpAdapterException $e) {
-            $response = (string) $e->getResponse()->getBody();
+            $method = $this->method;
+
+            $response = $client->$method($this->urlAccessToken(), [
```
rtheunissen: Felt like cleaning this up a bit, removed the `switch`.

src/Provider/AbstractProvider.php
```diff
@@ -295,24 +277,16 @@ public function getAccessToken($grant = 'authorization_code', array $params = []
...
-     * Creates a PSR-7 compatible request instance that can be modified.
-     * Often used to create calls against an API that requires authentication.
-     *
-     * @param  string $method
-     * @param  string $url
-     * @param  AccessToken $token
-     * @return RequestInterface
      */
-    public function getAuthenticatedRequest($method, $url, AccessToken $token)
+    public function getRequest($method, $url, AccessToken $token, $body = null)
```
rtheunissen: Figured a provider request might not be authorized if the access token is null, or if it's considered to always be authorized then why not call it `getRequest` to be consistent with `getResponse`.

ramsey:
I don't understand how this differs from #315. Additionally, there are conflicts, so I can't merge it.

rtheunissen:
Fair enough, would have made more sense to fork #315 instead.

PR #319, opened 2015-06-04
Add hasExpired to AccessToken

rtheunissen:
Seemed like an easy value-add to the token, even though it doesn't technically solve a problem. Makes it easy to check if a token should be refreshed when a request is made.

### DIFF
--- a/src/Token/AccessToken.php
+++ b/src/Token/AccessToken.php
@@ -100,6 +100,21 @@ class AccessToken
     }
 
     /**
+     * Checks if the token has expired.
+     *
+     * @return boolean true if the token has expired, false otherwise.
+     * @throws InvalidArgumentException if the expires is not set on the token.
+     */
+    public function hasExpired()
+    {
+        if (!isset($this->expires)) {
+            throw new InvalidArgumentException('"expires" is not set on the token');
+        }
+
+        return $this->expires < time();
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


ramsey:
Looks good. Thanks!

src/Token/AccessToken.php
```diff
@@ -100,6 +100,21 @@ public function getUid()
...
     /**
+     * Checks if the token has expired.
+     *
+     * @return boolean true if the token has expired, false otherwise.
+     * @throws InvalidArgumentException if the expires is not set on the token.
+     */
+    public function hasExpired()
+    {
+        if (!isset($this->expires)) {
+            throw new InvalidArgumentException('"expires" is not set on the token');
```
shadowhand: I'm not sure that using `InvalidArgumentException` is the best choice here, because there is no argument at play. A better choice would be `RuntimeException`.

ramsey: I gave that some thought, too, but I wasn't sure what would best fit. I think you're right: `RuntimeException` would be best.

rtheunissen: I thought about this too when I was writing it and went through all descriptions for them. I figured that it's an `InvalidArgumentException` because 'expires' is an invalid argument because it's not set. Another thought was to not throw and exception at all, but return false instead if not set.

PR #321, opened 2015-06-05
Token expires from InvalidArgumentException to RuntimeException

rtheunissen:
As per comments on #319


ramsey:
Looks great. Thanks for making these changes.

PR #322, opened 2015-06-05
Fix documentation about Refreshing a Token

ZenCocoon:


ramsey:
In the context of this example, I think there is no `$token` object on which a `refreshToken` property exists. `$refreshToken` is a variable itself, since it could have come from a data source (like a database) where you stored the refresh token. Maybe just add some copy above the example to note that the `$refreshToken` comes from some data store where we stored the token after earlier requests?

ZenCocoon:
Good point, how do you feel about this explanation instead then?

ramsey:
Looks great. Thanks!

PR #325, opened 2015-06-07
Add sendRequest between getRequest and parseRequest

rtheunissen:
Currently a request is sent and parsed in the same method, which makes it hard for anything that would like to get access to the response instance before it gets parsed. Making it protected allows an extending class (such as an API wrapper) to access the `ResponseInterface` without changing the public API of the provider.

The use case I came across was to log the status code of each response.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -354,6 +354,26 @@ abstract class AbstractProvider implements ProviderInterface
     }
 
     /**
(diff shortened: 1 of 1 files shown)


shadowhand:
I like it.

rtheunissen:
@ramsey Could you please take a look at this sometime? I would like to add this to the impending Guzzle 6 branch but would like to know if we like it or not before I spend any time on it.

ramsey:
LGTM

PR #326, opened 2015-06-10
If there is an error parsing a JSON response set the response as the error.

tenfef:
When connecting to a shitty API that returns a text error (or an HTML error)  even though we are expecting JSON, then we should escalate the text error otherwise all you get is "Unknown Error" which is super unhelpful.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -232,6 +232,8 @@ abstract class AbstractProvider implements ProviderInterface
 
                 if (JSON_ERROR_NONE === json_last_error()) {
                     $result = $json;
+                } elseif(json_last_error()){                    
+                    $result['error'] = $response;
                 }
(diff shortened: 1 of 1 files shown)


ramsey:
Please take a look at the 1.0 branch. I think this has already been corrected there.

shadowhand:
It has indeed.

shadowhand:
Further, this PR can't be done without a minor version bump, since it would change the API.

tenfef:
Thanks for the fast response guys, will check out 1.0 (I was not aware of this branch)

rtheunissen:
I'm working on a branch which is blocked by #315 which handles JSON a bit nicer. If #315 is accepted and merged, it would mean that we require PHP 5.5, and that we can use `json_last_error_msg`, which should help on top of what's already implemented on 1.0

PR #327, opened 2015-06-10
Remove LinkedIn Default Scopes

stevenmaguire:
As per the conversation in #307, removing the default scopes from the LinkedIn Provider included in this package. As a result, I've updated the documentation to point out this significant change for the LinkedIn Provider. 

cc @ramsey


stevenmaguire:
The extra `welcome back facebook` commits were due to a lingering change from when we accidentally removed Facebook from the readme and then rolled back :coffee:

ramsey:
Thanks!

PR #328, opened 2015-06-11
Check if property `publicProfileUrl` exists in LinkedIn response

leandersikma:
There is a scenario where the publicProfileUrl is not set, which results in a PHP notice.


shadowhand:
:+1:

ramsey:
Thanks. Merging in, but also pinging @stevenmaguire to make sure this isn't in the external LinkedIn library.

stevenmaguire:
This does not appear to be a problem in the external library, which has a different implementation for null/not set. 

https://github.com/thephpleague/oauth2-linkedin/blob/master/src/Provider/LinkedIn.php#L80

@leandersikma feel free to pull the project and write a test to validate this against your scenario.

PR #329, opened 2015-06-11
Remove ProviderInterface

rtheunissen:
I asked about this on Gitter but didn't get any response.

> What is the benefit of having a ProviderInterface, as opposed to having only an AbstractProvider? When would anyone want to use the interface instead of the abstract class?


shadowhand:
:-1: regardless of having an abstract class, we always want an interface to verify the contract. Having an abstract is not sufficient.

rtheunissen:
> we always want an interface to verify the contract

I disagree, which is why I'm putting this up.

shadowhand:
You gain nothing by removing the interface, and you lose the ability to type hint _by interface only_ which is correct way to do dependency inversion. This is why we type hint against PSR7 interfaces, not Guzzle implementations of those interfaces.

rtheunissen:
When using an interface you are forced to have those required methods be public, when most if not all of them could and probably should be protected.

> This is why we type hint against PSR7 interfaces, not Guzzle implementations of those interfaces.

Yes that is very clear but I don't think it's a fair comparison. The PSR 7 interfaces are at the top, and have no predefined behaviour. They are pure interfaces. What we have, also at the top of the chain, is by nature a 90% complete OAuth2 provider, with a few gaps that need to be filled by extending providers. No one one ever implement the interface rather than extend the abstract, and you still go by contract if you type hint the abstract class.

shadowhand:
> No one one ever implement the interface rather than extend the abstract, and you still go by contract if you type hint the abstract class.

This is a good argument. I'll defer to @ramsey on this, though I still feel we might be losing something here.

rtheunissen:
Travis timed out again it seems. :snail:

ramsey:
I want to think about this a bit before making a decision.

rtheunissen:
> I want to think about this a bit before making a decision.

No rush at all, might be worth getting a few more opinions in here?

ramsey:
Okay, @rtheunissen, you do make a good argument. So, after some consideration, I'm merging this in. Thanks!

rtheunissen:
:sparkles:

ramsey:
That reference was an accident. I put in the wrong issue number. :-)

PR #331, opened 2015-06-11
Use "Content-Type" headers to parse response

rtheunissen:
There has been a few cases where relying on a static expected response type is problematic. @SammyK mentioned that Facebook is inconsistent between API versions, so we added a dynamic way to get the response type in #301.

This pull request takes that one step further, and determines the response type by looking at the content-type headers returned by the response. I believe that this is more reliable and allows handling of API's / hosts that send multiple content types (for example Reddit sending image/png for some endpoints).

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -7,6 +7,7 @@ use GuzzleHttp\Client as HttpClient;
 use GuzzleHttp\ClientInterface as HttpClientInterface;
 use GuzzleHttp\Exception\BadResponseException;
 use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
 use League\OAuth2\Client\Provider\Exception\IdentityProviderException;
 use League\OAuth2\Client\Grant\GrantFactory;
 use League\OAuth2\Client\Token\AccessToken;
@@ -17,16 +18,6 @@ use InvalidArgumentException;
 
 abstract class AbstractProvider implements ProviderInterface
 {
-    /**
-     * @var string JSON response type.
-     */
-    const RESPONSE_TYPE_JSON = 'json';
-
-    /**
-     * @var string Parameter string response type.
-     */
-    const RESPONSE_TYPE_STRING = 'string';
-
     /**
      * @var string Key used in the access token response to identify the user.
      */
@@ -350,10 +341,14 @@ abstract class AbstractProvider implements ProviderInterface
 
         $request  = $this->getRequest($method, $url, $options);
         $response = $this->getResponse($request);
+
+        $this->checkResponse($response);
+
         $response = $this->prepareAccessTokenResult($response);
 
         return $grant->handleResponse($response);
     }
+
 
     /**
      * Get a request instance.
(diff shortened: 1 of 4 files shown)


shadowhand:
Makes sense to me. I would suggest squashing your two commits into one.

rtheunissen:
Okay will do.

src/Provider/AbstractProvider.php
```diff
@@ -414,56 +409,75 @@ protected function sendRequest(RequestInterface $request)
...
-    protected function getResponseType()
+    protected function parseJson($content)
     {
-        return static::RESPONSE_TYPE_JSON;
+        $content = json_decode($content, true);
+
+        if (json_last_error() !== JSON_ERROR_NONE) {
+            throw new UnexpectedValueException(sprintf(
+                "Failed to parse JSON response: %s",
+                json_last_error_msg()
```
rtheunissen: PHP 5.5 allows us to use [`json_last_error_msg()`](http://php.net/manual/en/function.json-last-error-msg.php)

src/Provider/AbstractProvider.php
```diff
@@ -414,56 +409,75 @@ protected function sendRequest(RequestInterface $request)
...
-                    throw new UnexpectedValueException('Unable to parse client response');
-                }
-                break;
-            case static::RESPONSE_TYPE_STRING:
-                parse_str($response, $result);
-                break;
+        $content = (string) $response->getBody();
+        $type = $this->getContentType($response);
+
+        if (strpos($type, "json") !== false) {
```
rtheunissen: Using `strpos` is more robust as a content-type header could be something like `test/json` or `application/json; charset=utf-8` which makes exact string matches more difficult.

src/Provider/AbstractProvider.php
```diff
@@ -414,56 +409,75 @@ protected function sendRequest(RequestInterface $request)
...
     /**
-     * Parse the response, according to the provider response type.
+     * Returns the content type header of a response.
+     *
+     * @param ResponseInterface $response
+     * @return string Semi-colon separated join of content-type headers.
+     */
+    protected function getContentType(ResponseInterface $response)
+    {
+        return join(";", (array) $response->getHeader('content-type'));
```
rtheunissen: Cast to an array because `$reponse->getHeader()` returns **null** if the header doesn't exist. I would have expected an empty array but nope, is null.

src/Provider/AbstractProvider.php
```diff
@@ -414,56 +409,75 @@ protected function sendRequest(RequestInterface $request)
...
         }
 
-        $this->checkResponse($result);
+        if (strpos($type, "urlencoded") !== false) {
+            parse_str($content, $parsed);
+            return $parsed;
+        }
 
-        return $result;
+        return $content;
```
rtheunissen: Just return the raw content if it wasn't urlencoded or JSON.

test/src/Provider/AbstractProviderTest.php
```diff
@@ -507,4 +477,50 @@ public function testInvalidAccessTokenMethod()
...
+    public function getPrivateMethod($class, $name)
+    {
+        $class = new \ReflectionClass($class);
+        $method = $class->getMethod($name);
+        $this->assertFalse($method->isPublic());
+        $method->setAccessible(true);
+        return $method;
+    }
+
+    private function _testParse($body, $type, $expected = null)
```
rtheunissen: @shadowhand is there a nicer way to do this? Makes for nice, small parse test cases though. :)

test/src/Provider/Fake.php
```diff
@@ -48,7 +48,7 @@ protected function prepareUserDetails(array $response, AccessToken $token)
         return new Fake\User($response);
     }
 
-    protected function checkResponse(array $response)
+    protected function checkResponse($response)
```
rtheunissen: Could now be an array or a string so can't typehint `array`.

test/src/Provider/AbstractProviderTest.php
shadowhand: Dataprovider :)

rtheunissen: I was using one but wasn't sure how to handle the expected exception case? I'll do some reading today.

PR #332, opened 2015-06-12
Proposed cleanup of getAuthorizationUrl and getAccessToken

rtheunissen:
The main goal behind this pull request is to modularize some of our long methods. This allows for easier, more unit-like testing, and provides flexibility for extending providers as there are more access points along the pipeline. 

`getAuthorizationUrl` split into:
 - `urlAuthorize`
 - `getAuthorizationParameters`
 - `getAuthorizationQuery`

`getAccessToken` split into:
 - `verifyGrant`
 - `prepRequestParams` on `$grant`
 - `getAccessTokenRequest`
    - `urlAccessToken`
    - `getAccessTokenQuery`
    - `switch` on method
    - `getRequest`
 - `getResponse`
 - `prepareAccessTokenResponse`
 - `handleResponse` on `$grant`

### Notes

1. Removed `httpBuildQuery` in favour of `getAuthorizationQuery` and `getAccessTokenQuery`, both of which use basic `http_build_query` with default parameters by default. Providers can override this if required.

2. Removed default parameter value for `$grant`. It made no sense: default grant was "authorization_code" which is a good default value but in order to provide the `code` and `state` parameters you have to provide options, thereby having no choice but to specify "authorization_code" anyway. An alternative here is to swap the order of `$options` and `$grant`?

3. `prepareAccessTokenResult` renamed to `prepareAccessTokenResponse` for naming consistency.

4. Added `appendQuery(url, query)` to try keep query string appending consistent. This is used in `getAccessTokenRequest` and `getAuthorizationUrl`.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -259,7 +259,7 @@ abstract class AbstractProvider
         ];
 
         if (is_array($options['scope'])) {
-            $options['scope'] = join(static::SCOPE_SEPARATOR, $options['scope']);
+            $options['scope'] = implode(static::SCOPE_SEPARATOR, $options['scope']);
         }
 
         // Store the state, it may need to be accessed later.
@@ -298,7 +298,7 @@ abstract class AbstractProvider
         $params = $this->getAuthorizationParameters($options);
         $query  = $this->getAuthorizationQuery($params);
 
-        return $base.'?'.$query;
+        return $this->appendQuery($base, $query);
     }
 
     public function authorize(array $options = [], $redirectHandler = null)
@@ -312,6 +312,25 @@ abstract class AbstractProvider
         header('Location: ' . $url);
         exit;
         // @codeCoverageIgnoreEnd
+    }
+
+    /**
+     * Appends a query string to a URL.
+     *
+     * @param string $url The URL to append the query to
+     * @param string $query The HTTP query string
+     *
+     * @return string
+     */
+    protected function appendQuery($url, $query)
+    {
+        $query = trim($query, '?&');
+
+        if ($query) {
+            return $url.'?'.$query;
+        }
+
+        return $url;
     }
 
     /**
@@ -367,7 +386,7 @@ abstract class AbstractProvider
 
         switch ($method) {
             case 'GET':
-                $url .= $query;
+                $url = $this->appendQuery($url, $query);
                 break;
             case 'POST':
                 $options['body'] = $query;
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -420,6 +420,9 @@ abstract class AbstractProvider
         $params   = $grant->prepRequestParams($params, $options);
         $request  = $this->getAccessTokenRequest($params);
         $response = $this->getResponse($request);
+
+        $this->checkResponse($response);
+
         $response = $this->prepareAccessTokenResponse($response);
 
         return $grant->handleResponse($response);
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -69,12 +69,6 @@ abstract class AbstractProvider
     protected $randomFactory;
 
     /**
-     * @var int This represents: PHP_QUERY_RFC1738, which is the default value for php 5.4
-     *          and the default encoding type for the http_build_query setup
-     */
-    protected $httpBuildEncType = 1;
-
-    /**
      * @param array $options
      * @param array $collaborators
      */
@@ -243,11 +237,18 @@ abstract class AbstractProvider
      */
     abstract protected function getDefaultScopes();
 
-    public function getAuthorizationUrl(array $options = [])
+    /**
+     * Returns authorization parameters based on provided options.
+     *
+     * @param array $options
+     * @return array Authorization parameters
+     */
+    protected function getAuthorizationParameters(array $options)
     {
         if (empty($options['state'])) {
             $options['state'] = $this->getRandomState();
         }
+
         if (empty($options['scope'])) {
             $options['scope'] = $this->getDefaultScopes();
         }
@@ -258,13 +259,13 @@ abstract class AbstractProvider
         ];
 
         if (is_array($options['scope'])) {
-            $options['scope'] = implode(static::SCOPE_SEPARATOR, $options['scope']);
+            $options['scope'] = join(static::SCOPE_SEPARATOR, $options['scope']);
         }
 
         // Store the state, it may need to be accessed later.
         $this->state = $options['state'];
 
-        $params = [
+        return [
             'client_id'       => $this->clientId,
             'redirect_uri'    => $this->redirectUri,
             'state'           => $this->state,
@@ -272,8 +273,32 @@ abstract class AbstractProvider
             'response_type'   => $options['response_type'],
             'approval_prompt' => $options['approval_prompt'],
         ];
+    }
 
-        return $this->urlAuthorize().'?'.$this->httpBuildQuery($params, '', '&');
+    /**
+     * Builds the authorization URL's query string.
+     *
+     * @param array $params Query parameters
+     * @return string Query string
+     */
+    protected function getAuthorizationQuery(array $params)
+    {
+        return http_build_query($params);
+    }
+
+    /**
+     * Returns the authorization URL for given options.
+     *
+     * @param array $options
+     * @return string
+     */
+    public function getAuthorizationUrl(array $options = [])
+    {
+        $base   = $this->urlAuthorize();
+        $params = $this->getAuthorizationParameters($options);
+        $query  = $this->getAuthorizationQuery($params);
+
+        return $base.'?'.$query;
     }
 
     public function authorize(array $options = [], $redirectHandler = null)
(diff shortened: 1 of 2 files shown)


src/Provider/AbstractProvider.php
```diff
@@ -271,13 +272,13 @@ public function getAuthorizationUrl(array $options = [])
         ];
 
         if (is_array($options['scope'])) {
-            $options['scope'] = implode(static::SCOPE_SEPARATOR, $options['scope']);
+            $options['scope'] = join(static::SCOPE_SEPARATOR, $options['scope']);
```
shadowhand: `join` is an alias for `implode`, so we should always use `implode`.

src/Provider/AbstractProvider.php
```diff
@@ -285,8 +286,32 @@ public function getAuthorizationUrl(array $options = [])
...
+     * @param array $options
+     * @return string
+     */
+    public function getAuthorizationUrl(array $options = [])
+    {
+        $base   = $this->urlAuthorize();
+        $params = $this->getAuthorizationParameters($options);
+        $query  = $this->getAuthorizationQuery($params);
+
+        return $base.'?'.$query;
```
shadowhand: What if there is no query?

src/Provider/ProviderInterface.php
```diff
@@ -61,7 +61,7 @@ public function authorize(array $options = []);
      * @param array $params
      * @return AccessToken
      */
-    public function getAccessToken($grant = 'authorization_code', array $params = []);
+    public function getAccessToken($grant, array $params = []);
```
shadowhand: Doesn't this mean that we can typehint against `GrantInterface`?

src/Provider/AbstractProvider.php
```diff
@@ -312,35 +337,53 @@ protected function getAccessTokenMethod()
...
+        $query = $this->getAccessTokenQuery($params);
         $method = strtoupper($this->getAccessTokenMethod());
+
         $options = [];
 
         switch ($method) {
             case 'GET':
-                // No providers included with this library use get but 3rd parties may
-                $url .= (parse_url($url, PHP_URL_QUERY) ? '&' : '?') . $requestParams;
+                $url .= $query;
```
shadowhand: What happened to the `?` or `&` joiner?

src/Provider/AbstractProvider.php
rtheunissen: Can you please explain why we *should*?

src/Provider/AbstractProvider.php
rtheunissen: Good point. I thought about it but figured it's not likely that there is a case where no query string is present in the auth url. How else would it work? Even then, an empty query string (or trailing ?) wouldn't affect the URL.

src/Provider/ProviderInterface.php
rtheunissen: We are currently accepting either string or `GrantInterface`, having the `GrantFactory` take care of that for us. Doesn't seem nice to *require* and instantiated `GrantInterface`.

src/Provider/AbstractProvider.php
rtheunissen: It's redundant as you can append a query string by overriding `getAccessTokenQuery` rather than include it in your base access token URL.

src/Provider/AbstractProvider.php
```diff
@@ -312,35 +341,53 @@ protected function getAccessTokenMethod()
...
+     * @param mixed $grant
+     * @return GrantInterface
+     */
+    protected function verifyGrant($grant)
     {
         if (is_string($grant)) {
-            $grant = $this->grantFactory->getGrant($grant);
-        } else {
-            $this->grantFactory->checkGrant($grant);
+            return $this->grantFactory->getGrant($grant);
```
shadowhand: @rtheunissen but if the grant is a string, it gets turned into an instance here.

src/Provider/AbstractProvider.php
```diff
@@ -312,35 +341,53 @@ protected function getAccessTokenMethod()
...
+    }
 
+    /**
+     * Returns a prepared request for requesting an access token.
+     *
+     * @param array $params Query string parameters
+     */
+    protected function getAccessTokenRequest(array $params)
+    {
         $url = $this->urlAccessToken();
```
shadowhand: This probably won't include a `?` query part in most cases.

src/Provider/AbstractProvider.php
```diff
@@ -312,35 +341,53 @@ protected function getAccessTokenMethod()
...
+        $query = $this->getAccessTokenQuery($params);
```
shadowhand: This currently doesn't add a `?` at all.

src/Provider/AbstractProvider.php
shadowhand: So this will currently produce an invalid URI: `http://example.com/auth/tokentype=access_code&etc=`

rtheunissen: Why would the provider's base url end on `&etc=`?

src/Provider/AbstractProvider.php
rtheunissen: You're right, I have fixed this in 4e0251110571ad7841d87dbe74f0f8d91383a32f. Wondering if there's a better way that we can test that as the tests pass either way.

src/Provider/AbstractProvider.php
rtheunissen: Correct. Providers shouldn't need to have a query string in the base url methods. If for some reason they *need* to, they can override `getAccessTokenQuery` or just specify it as an option I think? It would go through the grant, and end up being passed to `getAccessTokenQuery` anyway. 

Policy: No query string in base url, there are multiple other better alternatives.

src/Provider/AbstractProvider.php
rtheunissen: Fixed this in 76333a727bfc55ec6053b0e3760a44bd09b6a5b4

src/Provider/AbstractProvider.php
rtheunissen: Fixed in 44be55907d03c4cfa3d73e3b35144838a2e536b5

src/Provider/AbstractProvider.php
shadowhand: Because the alias only exists for people that are confused, and we are not confused.

test/src/Provider/AbstractProviderTest.php
```diff
@@ -507,4 +506,32 @@ public function testInvalidAccessTokenMethod()
...
+
+    private function getMethod($class, $name)
+    {
+        $class = new \ReflectionClass($class);
+        $method = $class->getMethod($name);
+        $method->setAccessible(true);
+        return $method;
+    }
+
+    public function getAppendQueryProvider()
```
shadowhand: I think the rest of the data providers use the form `providerNameOfTest` or similar, with `provider` always used as the prefix.

src/Provider/AbstractProvider.php
rtheunissen: Hah okay, will revert. Thanks

test/src/Provider/AbstractProviderTest.php
rtheunissen: Okay thanks good to know. Fixed in 22fee7b94605ae5703eaaecd1136f3296166596a

src/Provider/AbstractProvider.php
ramsey: Where is 3da0cdd? It doesn't appear to be added to this pull request.

src/Provider/AbstractProvider.php
ramsey: Again, I'm not seeing your fix in this pull request.

src/Provider/AbstractProvider.php
ramsey: Another case where I don't see the fix in this pull request. :-)

src/Provider/AbstractProvider.php
ramsey: Where is 44be559?

ramsey:
Looks like the pull request might be missing some commits, or maybe they were made in other pull requests that I merged in before this one? If so, please rebase your branch against the most recent 1.0 branch.

src/Provider/AbstractProvider.php
rtheunissen: I think I made a big mistake by rebasing on 1.0.

rtheunissen:
I did this:

```bash
git fetch upstream
git rebase 1.0
# fixed some conflicts, continued
git push origin cleaner_build_query -f
```

Did I make a mistake?

shadowhand:
@rtheunissen it looks like you used your local `1.0` branch, since you didn't specify `git rebase upstream/1.0`.

ramsey:
If you only fetched, then your local `1.0` branch didn't have the changes in it, so when you rebased, you weren't getting the recent changes.

Try `git rebase upstream/1.0` to get the changes into your working branch.

ramsey:
What @shadowhand said. :-)

rtheunissen:
@shadowhand I also did `git checkout 1.0 && git pull upstream 1.0`? Assumed that would take care of it. My bad. :/ What's the best thing to do now?

rtheunissen:
image

ramsey:
@shadowhand @rtheunissen While I have both of your attentions here, once this is merge in, any objections to me merging `1.0` into `master`, setting `master` as the branch alias for `1.0.0-dev`, and pointing the default branch on GitHub to the `0.12.x` branch?

rtheunissen:
Sounds good to me, no objections.

PR #333, opened 2015-06-12
Class name resolution as scalar with "class" keyword

rtheunissen:
A proposed change from explicit string class names to namespaced resolution, meaning if we want to change the class we are using, we only need to `use` a different class and alias accordingly. This is only possible now that we're on version >= 5.5 for the 1.0 branch.

Eg.

`'Psr\Http\Message\StreamInterface` --> `StreamInterface::class`.

### DIFF
--- a/src/Grant/GrantFactory.php
+++ b/src/Grant/GrantFactory.php
@@ -65,7 +65,7 @@ class GrantFactory
      */
     public function isGrant($class)
     {
-        return is_subclass_of($class, 'League\OAuth2\Client\Grant\GrantInterface');
+        return is_subclass_of($class, GrantInterface::class);
     }
 
     /**
--- a/test/src/Provider/AbstractProviderTest.php
+++ b/test/src/Provider/AbstractProviderTest.php
@@ -2,11 +2,21 @@
 
 namespace League\OAuth2\Client\Test\Provider;
 
-use GuzzleHttp\Exception\BadResponseException;
 use League\OAuth2\Client\Provider\AbstractProvider;
-use League\OAuth2\Client\Token\AccessToken;
-use League\OAuth2\Client\Provider\Exception\IdentityProviderException;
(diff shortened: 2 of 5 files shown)


src/Grant/GrantFactory.php
```diff
@@ -65,7 +65,7 @@ protected function registerDefaultGrant($name)
      */
     public function isGrant($class)
     {
-        return is_subclass_of($class, 'League\OAuth2\Client\Grant\GrantInterface');
+        return is_subclass_of($class, GrantInterface::class);
```
shadowhand: This makes sense, we already have the `GrantInterface` imported.

shadowhand:
I'm not entirely convinced about using `::class` inside of tests, it feels like change for the sake of change.

test/src/Provider/AbstractProviderTest.php
```diff
@@ -245,10 +255,10 @@ public function testRandomGeneratorCreatesRandomState()
     {
         $xstate = str_repeat('x', 32);
 
-        $generator = m::mock('RandomLib\Generator');
+        $generator = m::mock(RandomGenerator::class);
         $generator->shouldReceive('generateString')->with(32)->times(1)->andReturn($xstate);
 
-        $factory = m::mock('RandomLib\Factory');
+        $factory = m::mock(randomFactory::class);
```
ramsey: The "r" should be capitalized in `RandomFactory`

rtheunissen: > I'm not entirely convinced about using ::class inside of tests, it feels like change for the sake of change.

@shadowhand not at all. If we want to change the namespace for something, we don't have to refactor the entire file. Less code, without sacrificing clarity, which is easier to maintain: definitely not for the sake of change.

ramsey:
I think this is fine, and since it's mostly touching tests, it's doesn't present any risk.

PR #336, opened 2015-06-14
Set correct scope separator for Instagram

rtheunissen:
Instagram's scope separator is `+`, as documented [here](https://instagram.com/developer/authentication/#scope). This patch fixes that on `master`.

### DIFF
--- a/src/Provider/Instagram.php
+++ b/src/Provider/Instagram.php
@@ -6,6 +6,7 @@ use League\OAuth2\Client\Entity\User;
 
 class Instagram extends AbstractProvider
 {
+    public $scopeSeparator = '+';
     public $scopes = ['basic'];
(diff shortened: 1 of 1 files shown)


shadowhand:
Did you also make an issue on [oauth2-instagram](https://github.com/thephpleague/oauth2-instagram)?

rtheunissen:
Didn't realise they were separate packages now. Not sure how I missed that.

rtheunissen:
@shadowhand Created an issue [here](https://github.com/thephpleague/oauth2-instagram/issues/1)

shadowhand:
No need to close this, the split provider is only for 1.0+. Would be good to fix this one master too. :)

rtheunissen:
I'm clearly having a good day it seems.

PR #339, opened 2015-06-16
Fixing provider state in the README

patrickconroy:
It's a protected variable, users should use `getState()` instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ if (!isset($_GET['code'])) {
 
     // If we don't have an authorization code then get one
     $authUrl = $provider->getAuthorizationUrl();
-    $_SESSION['oauth2state'] = $provider->state;
+    $_SESSION['oauth2state'] = $provider->getState();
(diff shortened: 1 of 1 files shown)


patrickconroy:
Never mind, I didn't realize how different master and the ~0.1 branches are.

patrickconroy:
After looking at it again for a 5th time, this change looks like it should be made. Master's `AbstractProvider->state` is protected, and won't work like the docs say.

shadowhand:
:+1: Looks right.

shadowhand:
@patrickconroy the master branch has just replaced 1.0

ramsey:
There is no `getState()` method on the providers in `0.12.x`, and the `$state` property is public, as well. It's only in `master` (i.e. 1.0.0-dev) that this changes. To avoid confusion, please open a new pull request against master to make this change to the README there.

We'll leave the README in `0.12.x` as is.

PR #341, opened 2015-06-16
Added support for JSONP responses

SammyK:
Most of Facebook responses are of type [JSONP](https://en.wikipedia.org/wiki/JSONP) which have a [content-type header of `application/javascript`](http://stackoverflow.com/questions/477816/what-is-the-correct-json-content-type#answers-header). This PR adds support to properly decode JSONP responses.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -537,7 +537,9 @@ abstract class AbstractProvider
         $content = (string) $response->getBody();
         $type = $this->getContentType($response);
 
-        if (strpos($type, "json") !== false) {
+        // JSON  = application/json
+        // JSONP = application/javascript
+        if (strpos($type, "json") !== false || strpos($type, "javascript") !== false) {
             return $this->parseJson($content);
         }
 
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -522,7 +522,7 @@ abstract class AbstractProvider
      */
     protected function getContentType(ResponseInterface $response)
     {
-        return join(";", (array) $response->getHeader('content-type'));
+        return join(';', (array) $response->getHeader('content-type'));
     }
 
     /**
@@ -539,11 +539,11 @@ abstract class AbstractProvider
 
         // JSON  = application/json
         // JSONP = application/javascript
-        if (strpos($type, "json") !== false || strpos($type, "javascript") !== false) {
+        if (strpos($type, 'json') !== false || strpos($type, 'javascript') !== false) {
             return $this->parseJson($content);
         }
 
-        if (strpos($type, "urlencoded") !== false) {
+        if (strpos($type, 'urlencoded') !== false) {
             parse_str($content, $parsed);
             return $parsed;
         }


ramsey:
I don't believe JSONP is parsable JSON, is it? I thought it was JSON wrapped in a JavaScript function. For example:

```
myFunc({"foo": "bar"})
```

Our JSON parser won't be able to parse this as JSON, unless I'm mistaken.

src/Provider/AbstractProvider.php
```diff
@@ -537,7 +537,9 @@ protected function parseResponse(ResponseInterface $response)
         $content = (string) $response->getBody();
         $type = $this->getContentType($response);
 
-        if (strpos($type, "json") !== false) {
+        // JSON  = application/json
+        // JSONP = application/javascript
+        if (strpos($type, "json") !== false || strpos($type, "javascript") !== false) {
```
shadowhand: Since you're working on this, can you change these double quotes to be single? Just a pet peeve of mine that I didn't notice when it was changed the first time around.

ramsey: :+1:

SammyK:
@ramsey Facebook just returns a normal JSON string with the `application/javascript` content-type header.

shadowhand:
what

ramsey:
I'm still wary of treating `application/javascript` as parsable JSON. That may be how Facebook does it, but we can't guarantee it will always work, and we have no way to bail out if someone else sends `application/javascript` and it's proper JavaScript (not parsable JSON).

Is there a way to handle this in the Facebook client library, rather than here in the base library?

shadowhand:
There totally is:

```php
protected function parseResponse(ResponseInterface $response)
{
    $response = $response->withHeader('Content-Type', 'application/json');
    return parent::parseResponse($response);
}
```

SammyK:
@ramsey Yeah, I can just overwrite `getContentType()` with a fix. :)

ramsey:
Awesome. Thanks!

rtheunissen:
@shadowhand `withHeader` may be overkill as it creates a complete duplicate of the original instance. Why not

```php
protected function getContentType(ResponseInterface $response)
{   
    $type = parent::getContentType($response);
    
    if (strpos($type, 'application/javascript') !== false) {
        return 'application/json';
    }
}
```

@SammyK I'm assuming this is close to what you meant by `overwrite getContentType() with a fix`?

SammyK:
@rtheunissen I already fixed it [similarly to how you proposed](https://github.com/thephpleague/oauth2-facebook/blob/master/src/Provider/Facebook.php#L142-L160). Facebook also returns url-encoded key/value pair responses as `text/plain`... we need another dog gif here.. :)

rtheunissen:
@SammyK Very nice :+1:

SammyK:
@rtheunissen I agree. I like your way better. Done! :)

rtheunissen:
:sparkles:

SammyK:
Sent a [bug report to Facebook](https://developers.facebook.com/bugs/1634595546786003/).

shadowhand:
@rtheunissen @SammyK oops I didn't notice that `getContentType` is a method of the `AbstractProvider`. :blush: I read it as a convenience method of the Guzzle `Request`.

SammyK:
In case anyone else is having issues with this on Facebook's end, they have confirmed it as a bug and have a [fix in the queue](https://developers.facebook.com/bugs/1634595546786003/).

PR #342, opened 2015-06-16
Add mention of PSR-7 to readme

SammyK:
Since the package is using the latest Guzzle release, it's all PSR-7y now. :)


ramsey:
Is PSR-7 something that a package can be "compliant with?" It's not like it's a coding standard for us to adopt. Underlying packages may adhere to the PSR-7 interfaces, but I don't understand how that's important to this package.

shadowhand:
Maybe "makes use of" would be more appropriate here. But I do think we should market the fact that the client works with PSR-7 interfaces, because it allows a lot more flexibility when creating APIs on top of the providers.

SammyK:
> But I do think we should market the fact that the client works with PSR-7 interfaces, because it allows a lot more flexibility when creating APIs on top of the providers.

That was my thought. :)

ramsey:
Okie dokie.

PR #343, opened 2015-06-16
Single quotes in the context of content types

rtheunissen:
As mentioned and then orphaned in #341, this fixes some double quotes that should be single quotes for the sake of consistency.


shadowhand:
:+1: I'm of the opinion that single quotes should always be used unless interpolation is explicitly needed.

rtheunissen:
Same here, not sure why I used double in this instance. Thanks for pointing it out earlier.

PR #344, opened 2015-06-16
Remove GrantInterface

rtheunissen:
This is a very similar case to #329, where the interface is arguably redundant. Here is my take on combining interfaces with abstract classes.

**Interface + Abstract**
This is when you rely on the interface to enforce a contract, and it is plausible for an instance to implement the interface directly. The interface itself has value even though it doesn't provide any default or suggested implementation details. By building an abstract class on top of the interface, you can make it easier for another implementation by taking care of *most* of an implementation in a recommended or arguably "default" way. A good example is a Cache which has `store`, `fetch` and `contains`. An abstract cache could take care of things like "before fetch, check if the cache `contains`, otherwise throw an exception" and add a `doFetch`. By extending the abstract, you inherit those safety checks, even though they have nothing to do with the interface.

**Interface only**
This is when it makes no sense to have any default or suggestion implementation at all. PSR's are a good example.

**Abstract only**
When the interface by itself doesn't really have much value other than being a contract enforces for the abstract class (I think this is what @shadowhand was saying). But an abstract class is also a contract enforcer. It's up to the maintainer to be aware of the contract defined by an abstract class. In the case of the `GrantInterface`, just like the `ProviderInterface`, it makes no sense to implement the interface directly. `AbstractGrant` provides the base functionality we want, but still enforces the contract we need (which in this case is to get required parameters and a string representation).

---

You could argue that it doesn't hurt to leave the interface alone, and I wouldn't disagree with that. But my argument is that it doesn't add any useful value either. ¯\\_(ツ)_/¯


ramsey:
I was attempting to resolved the conflicts when merging this in, but it seems there are significant changes between this PR and the nomenclature one, and I'd prefer if you resolved the conflicts (or created a new pull request), please.

Thanks!

rtheunissen:
Okay I think I rebased correctly this time.

ramsey:
Looks good. Thanks!

PR #345, opened 2015-06-16
A push for consistent nomenclature

rtheunissen:
> "The devising or choosing of names for things, especially in a science or other discipline."

I'm a firm believer that a consistent codebase is an intuitive codebase. This pull request is not important, and doesn't add any functionality; it's simply an attempt to improve the consistency of our naming.

**AbstractGrant**
- `getRequiredRequestParams` -> `getRequiredRequestParameters`
- `prepRequestParams ` -> `prepareRequestParameters `
- `handleResponse` -> `createAccessToken`

**AbstractProvider**
- `urlAuthorize` -> `getBaseAuthorizationUrl`
- `urlAccessToken` -> `getBaseAccessTokenUrl`
- `urlUserDetails` -> `getUserDetailsUrl`

These are not "change for the sake of change". You should be able to determine what a method does based on its name. `isValue`, `hasValue`, `setValue`, `getValue` etc are all staples.

`urlAuthorize` is internal, so could argue that it doesn't matter. But it always matters, even if it's private. It's not clear what `urlAuthorize` does; maybe attempt to authorize using a URL? I know it's a nitpick but if the goal is to write a solid, top shelf library, these are the kind of things that make a difference.

### DIFF
--- a/src/Grant/AbstractGrant.php
+++ b/src/Grant/AbstractGrant.php
@@ -20,19 +20,19 @@ abstract class AbstractGrant implements GrantInterface
      *
      * @return array
      */
-    abstract protected function getRequiredRequestParams();
+    abstract protected function getRequiredRequestParameters();
 
-    public function prepRequestParams(array $defaultParams, array $params)
+    public function prepareRequestParameters(array $defaults, array $params)
     {
         $this->checkRequiredParameters(
-            $this->getRequiredRequestParams(),
+            $this->getRequiredRequestParameters(),
             $params
         );
 
-        return array_merge($defaultParams, $params);
+        return array_merge($defaults, $params);
     }
 
-    public function handleResponse(array $response)
+    public function createAccessToken(array $response)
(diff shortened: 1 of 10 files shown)


ramsey:
I'll read through the changes tomorrow, but in general, I'm in favor of these name changes. They were on my mental list of things to do for 1.0.

:+1:

rtheunissen:
Thanks @ramsey

shadowhand:
This isn't a bad idea. Here's another suggestion for consideration: [Getters and Setters are Evil](http://www.yegor256.com/2014/09/16/getters-and-setters-are-evil.html).

ramsey:
The author does have some valid points about the procedural mindset vs. the object mindset, but as you said, getters/setters are a well-established pattern in our community. It's easy to grasp the semantics of getters/setters, and the previous names of these methods were not so easily understood, so I think these changes are for the best.

Issue #335, opened 2015-06-14
Instagram Multiple scopes


mrgodhani:
Hi

How would I set multiple scopes for Instagram provider ?

This is what I have :

```
'instagram' => [
        'clientId'      => env('INSTAGRAM_CLIENTID'),
        'clientSecret'  => env('INSTAGRAM_CLIENTSECRET'),
        'redirectUri'   => env('INSTAGRAM_REDIRECTURI'),
        'scopes'        => ['likes','basic']
    ]
```

And this is the error I get :
```
{"code": 400, "error_type": "OAuthException", "error_message": "Invalid scope field(s): likes,basic"}
```

rtheunissen:
I think the Instagram scope separator should be `+`. See [docs](https://instagram.com/developer/authentication/#scope)

rtheunissen:
PR to fix this has been opened [here](https://github.com/thephpleague/oauth2-client/pull/336). @ramsey

rtheunissen:
@mrgodhani I have opened an issue on [thephpleague/oauth2-instagram](https://github.com/thephpleague/oauth2-instagram/issues)

ramsey:
The scope separator for Instagram has been fixed in 0.12.0. Please let me know if that fixes this issue.

mrgodhani:
:+1:  I will check it out today and let you know

mrgodhani:
```
{"code": 400, "error_type": "OAuthException", "error_message": "Invalid scope field(s): basic+likes+comments+relationships"}
```
Still get this error

stevenmaguire:
I will take care of this and submit a PR.

stevenmaguire:
I've submitted a PR to correct the scope separator. Luckily I found that the solution also applies to the LinkedIn provider. I've submitted a separate PR for that. 

Additionally, both of these providers have been updated in their broken out packages.

ramsey:
The fix for this has been released in 0.12.1.

PR #346, opened 2015-06-17
Updating scope sepearator and adding test to Instagram provider

stevenmaguire:
As per #335 

I've added a test to ensure the resulting authorization url is formatted correctly with the scope separator being a space.

### DIFF
--- a/src/Provider/Instagram.php
+++ b/src/Provider/Instagram.php
@@ -6,7 +6,7 @@ use League\OAuth2\Client\Entity\User;
 
 class Instagram extends AbstractProvider
 {
-    public $scopeSeparator = '+';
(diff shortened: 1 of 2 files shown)


shadowhand:
Heh, of course... URL encoding changes ` ` to `+`. :+1:

stevenmaguire:
:wink: yep

Also, I did not realize the `scopeSeparator` property was being utilized, so I was setting an expectation to use a single index with all scopes space delimited, which worked fine. This is much better.

stevenmaguire:
This build timed out again, which caused the build error.

ramsey:
I'm restarting the job that timed out.

stevenmaguire:
Thanks @ramsey! It passed!

PR #347, opened 2015-06-17
Updating scope sepearator and adding test to LinkedIn provider

stevenmaguire:
Related to issue #335 and #346 , the LinkedIn provider can support scopes as a complete array, instead of a single index with all scopes space delimited. 

This also includes a test to ensure the authorization url is formatted correctly. 

This change has already be made and published in the separate package.


shadowhand:
:+1:

PR #348, opened 2015-06-17
Use Guzzle/Service/Client get Method to Set Header

Jaspaul:
Remove calls to `setDefaultOption` to set the headers for the
get request being generated by the AbstractProvider. This change
was made to fix `Requires authentication` errors being recieved
from GitHub and `Unknown authentication scheme` errors being
recieved from LinkedIn.

Looking into the documentation on Guzzle the get method on the
client will manage and set our headers for us, and was used in
place of `setDefaultOption`.

Tests were updated to reflect the removal of the call to
`setDefaultOption`.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -356,12 +356,7 @@ abstract class AbstractProvider implements ProviderInterface
         try {
             $client = $this->getHttpClient();
             $client->setBaseUrl($url);
-
-            if ($headers) {
-                $client->setDefaultOption('headers', $headers);
-            }
-
-            $request = $client->get()->send();
+            $request = $client->get(null, $headers)->send();
             $response = $request->getBody();
         } catch (BadResponseException $e) {
             // @codeCoverageIgnoreStart
--- a/test/src/Provider/EventbriteTest.php
+++ b/test/src/Provider/EventbriteTest.php
@@ -82,7 +82,6 @@ class EventbriteTest extends \PHPUnit_Framework_TestCase
         $client->shouldReceive('setBaseUrl')->times(5);
         $client->shouldReceive('post->send')->times(1)->andReturn($postResponse);
         $client->shouldReceive('get->send')->times(4)->andReturn($getResponse);
-        $client->shouldReceive('setDefaultOption')->times(4);
         $this->provider->setHttpClient($client);
 
         $token = $this->provider->getAccessToken('authorization_code', ['code' => 'mock_authorization_code']);
--- a/test/src/Provider/GithubTest.php
+++ b/test/src/Provider/GithubTest.php
@@ -121,7 +121,6 @@ class GithubTest extends \PHPUnit_Framework_TestCase
 
         $client = m::mock('Guzzle\Service\Client');
         $client->shouldReceive('setBaseUrl')->times(5);
-        $client->shouldReceive('setDefaultOption')->times(4);
         $client->shouldReceive('post->send')->times(1)->andReturn($postResponse);
         $client->shouldReceive('get->send')->times(4)->andReturn($getResponse);
         $this->provider->setHttpClient($client);
@@ -181,7 +180,6 @@ class GithubTest extends \PHPUnit_Framework_TestCase
 
         $client = m::mock('Guzzle\Service\Client');
         $client->shouldReceive('setBaseUrl')->times(2);
-        $client->shouldReceive('setDefaultOption')->times(1);
         $client->shouldReceive('post->send')->times(1)->andReturn($postResponse);
         $client->shouldReceive('get->send')->times(1)->andReturn($getResponse);
         $this->provider->setHttpClient($client);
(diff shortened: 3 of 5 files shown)


src/Provider/AbstractProvider.php
```diff
@@ -356,12 +356,7 @@ protected function fetchProviderData($url, array $headers = [])
         try {
             $client = $this->getHttpClient();
             $client->setBaseUrl($url);
```
shadowhand: Couldn't we also drop the `setBaseUrl` call here and use `get($url, $headers)` instead?

Jaspaul: Looks like we can, but I'm not sure how to update the tests to reflect this change. It looks like `setBaseUrl` is being called in other places, so I'm not sure what the expected number of calls to it will be.

Jaspaul: *What the number of expected calls **should** be.

ramsey: I honestly don't understand why our tests are asserting for the number of times a method is called internally. I think that makes for brittle tests, and we should probably just remove those lines.

ramsey:
@Jaspaul, is this something that will also need to be ported into the master branch for 1.0.0-dev?

Jaspaul:
@ramsey I don't think it applies to master. These changes are specific to `"guzzle/guzzle": "~3.7"` and master is running `"guzzlehttp/guzzle": "~6.0"`.

ramsey:
> This change was made to fix `Requires authentication` errors being recieved from GitHub and `Unknown authentication scheme` errors being recieved from LinkedIn.

@jaspaul, my question about whether this needs to be ported into the master branch was more to this point. Does 1.0.0-dev encounter these errors?

Additionally, does the [oauth2-linkedin](https://github.com/thephpleague/oauth2-linkedin) package have this problem? If so, this should be ported into that package.

Pinging @stevenmaguire, who maintains the oauth2-linkedin package.

stevenmaguire:
Wow! I missed this ping. Sorry. 

If this problem is not present in the version of the client running Guzzle 6, does this need to be addressed in the provider project? Maybe I am missing something. If so, @Jaspaul can you demonstrate the problem you are experiencing and create an issue in the [linkedin project](/thephpleague/oauth2-linkedin)?

ramsey:
I don't understand how setting the headers like this:

    $client->setDefaultOption('headers', $headers);
    $request = $client->get()->send();

is any different from this, in Guzzle 3.7:

    $request = $client->get(null, $headers)->send();

Can you show the problem so that we can reproduce it?

Thanks!

(At this point, I'm just assuming this is a bugfix for the 0.12.x series, and we will not need to port this into the 1.0.x series.)

stevenmaguire:
I spent a bit more time testing the LinkedIn provider today and I did uncover a small problem with the completing the access token retrieval. 

The access token request object needed to include a 'Content-Type: application/x-www-form-urlencoded` header to be successful with LinkedIn. I've included an override method in the LinkedIn provider that appends this header before completing the transaction.

@jaspaul, please pull the latest from the Linkedin provider and test this out to see if it alleviates your concerns. 

https://github.com/thephpleague/oauth2-linkedin/blob/master/src/Provider/LinkedIn.php#L131

@ramsey the spec does indicate that this header must be included for Access Token retrieval. It does not appear that the "stock" behavior current sets it. Should this change be rolled into the core package prior to beta? If so, I am happy to submit a PR and roll out the changes in the LinkedIn package.

http://tools.ietf.org/html/rfc6749#section-4.1.3

shadowhand:
@stevenmaguire I'm absolutely certain this was mentioned before and must have been forgotten. We absolutely should be sending a `Content-Type` header when appropriate.

stevenmaguire:
Good! I will prepare a PR as there doesn't seem to be one open.

rtheunissen:
This was discussed in #340 but I never got around to making it happen. @stevenmaguire I'm happy to do that right now, unless you are already?

stevenmaguire:
Oh! Sorry, @rtheunissen I was in the midst of putting the fix in. It is up at #381.

ramsey:
I've merged #381 into master. What should I do with this pull request? Merge into 0.12.x and tag a bugfix release there?

I'm still not sure how this change fixes the same content-type problem that #381 addresses for the 1.x series.

ramsey:
Is anyone able to answer my questions about this PR? How does it address the same issue that #381 fixes in 1.x? I see nothing in the diff dealing with `content-type`.

@Jaspaul, I'd like to merge in your pull request, so please help me understand the problem this is solving.

Thanks!

stevenmaguire:
I am not up to speed on the exact problem mentioned here. I only referenced the issue that was resolved in #381 in an effort to see if @Jaspaul felt that solution or the problem it solved would alleviate the symptoms reported here.

I have nothing further to contribute to this PR until more information is available.

ramsey:
@Jaspaul, anything?

ramsey:
Closing this ticket due to lack of activity.

@Jaspaul, if you're still having this problem and want this merged in, please open a new pull request and help me understand the problem it solves and how this code solves that problem. Thanks!

PR #350, opened 2015-06-20
Pass ResponseInterface and parsed body to checkResponse

### LINKED ISSUE #338: [1.0] Can't rely on BadRequestException
Guzzle 6 has an *optional* `http_errors` middleware, which means that it is not guaranteed that exceptions will be thrown for bad requests. 

I have two ideas that could help fix this:

1. 
 - Keep the `catch` for `BadRequestException` in order to use `$e->getReponse()`.
 - Check response status code and throw either a `ClientErrorException` or `ServerErrorException`, creating with the `Response`.
 - Could create protected `onError(request,response)` and `onSuccess(request, response)` style methods that by default throw those two exceptions.

2. Pass the `Response` to `checkResponse` *as well as* the parsed result. This would allow providers to determine errors based on status code and/or the parsed result.

---

Happy to PR this if it sounds reasonable to everyone.

rtheunissen:
Attempt to fix #338.

The problem that this is trying to solve:

It is not possible to detect an error when a service uses status codes to indicate failure rather than the response body. Currently, checkResponse only accepts the parsed body, so this PR proposes to also use the `ResponseInterface` alongside the parsed body. This allows a provider to utilise the status code or other metadata on the response to determine whether an error has occurred. 

Note: I've also added `checkResponse` to `getResponse` - not sure if that's what we want though? Wasn't sure if `checkResponse` is an access token request thing only.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -421,11 +421,9 @@ abstract class AbstractProvider
         $request  = $this->getAccessTokenRequest($params);
         $response = $this->getResponse($request);
 
-        $this->checkResponse($response);
+        $prepared = $this->prepareAccessTokenResponse($response);
 
-        $response = $this->prepareAccessTokenResponse($response);
-
-        return $grant->createAccessToken($response);
+        return $grant->createAccessToken($prepared);
     }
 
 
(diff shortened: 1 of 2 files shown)


ramsey:
It looks like `checkResponse()` is a generic way for the provider to implement its own error-checking capabilities, so I think this works out well.

Issue #352, opened 2015-06-20
Achieve 100% code coverage


ramsey:
We're so close to 100% covered in 1.0.0-dev, there's no reason we shouldn't make it all the way. There are just a few outlying lines not covered, and they shouldn't be difficult to test.

As of right now, here's what our coverage looks like:

![oauth2-client-coverage](https://cloud.githubusercontent.com/assets/42941/8268169/e952da1e-1740-11e5-87eb-173b0acb40b2.png)

ramsey:
We now have 100% code coverage.

PR #354, opened 2015-06-23
Bringing test coverage up to 100%

stevenmaguire:
In response to #352, I humbly submit these changes.

There is definitely some room to refactor some of the implementations to improve testability, and I think it is in ok shape here. 

One thing I want to point out.

#### Added a new fake provider to test AbstractProvider constant
The `AbstractProvider` includes a `ACCESS_TOKEN_UID` constant set to null. In order to completely test the `prepareAccessTokenResponse` method, this constant needed to be set. The choice I was presented with was to set the `Fake` class constant to a value, which broken many existing tests, or create a new child class of `Fake` where the constant is set. I chose the latter.

### DIFF
--- a/src/Token/AccessToken.php
+++ b/src/Token/AccessToken.php
@@ -108,11 +108,13 @@ class AccessToken
      */
     public function hasExpired()
     {
-        if (!isset($this->expires)) {
+        $expires = $this->getExpires();
+
+        if (empty($expires)) {
             throw new RuntimeException('"expires" is not set on the token');
         }
 
-        return $this->expires < time();
+        return $expires < time();
     }
 
     /**
--- a/src/Tool/MacAuthorizationTrait.php
+++ b/src/Tool/MacAuthorizationTrait.php
@@ -32,13 +32,20 @@ trait MacAuthorizationTrait
     // AbstractProvider
     abstract protected function getRandomState($length);
 
+    /**
+     * Get authorization headers
+     *
+     * @param  AccessToken $token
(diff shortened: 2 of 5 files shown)


rtheunissen:
> Added sleep to test AccessToken expiration check

Could you mock the response of `getExpires`, and change direct access to `$this->expires` to `$this->getExpires()`? That would be one way of getting past `sleep`. I'm not sure how else you could test this, with the property being protected.

stevenmaguire:
@rtheunissen that's a great idea. I did not notice the `getExpires` method as it was not being utilized by the `hasExpired` method. I've updated the solution to use this mock and updated the `AccessToken` class to utilize the `getExpires` method.

rtheunissen:
@stevenmaguire I wrote the `hasExpired` method and didn't think about test coverage at the time. Was thinking it's an unnecessary method call to access a property I already have access to. I'm not sure how I feel about adjusting code in order to test it. Another alternative, which doesn't affect the source, is to [use Reflection](http://stackoverflow.com/questions/18558183/phpunit-mockbuilder-set-mock-object-internal-property). 

My opinion is that if the tests can run quickly, and *test* that things work and behave, then I'm all for using things like Reflection.

stevenmaguire:
@rtheunissen I don't have much preference for either solution at the moment so I am happy to go with either if you have strong feelings and want to address it now. My goal here was to add test coverage :smile:

ramsey:
Thanks!

PR #355, opened 2015-06-24
Added AccessTokenInterface

lightningspirit:
Added AccessTokenInterface. Providers can now override AccessToken and add their own information.
This is really important since some idps providers other information and properties for the AccessToken.

### DIFF
--- a/src/Grant/AuthorizationCode.php
+++ b/src/Grant/AuthorizationCode.php
@@ -2,7 +2,7 @@
 
 namespace League\OAuth2\Client\Grant;
 
-use League\OAuth2\Client\Token\AccessToken;
+use League\OAuth2\Client\Token\AccessTokenInterface;
 
 class AuthorizationCode implements GrantInterface
 {
@@ -20,8 +20,8 @@ class AuthorizationCode implements GrantInterface
         return array_merge($defaultParams, $params);
     }
 
-    public function handleResponse($response = [])
+    public function handleResponse(AccessTokenInterface $token, array $response = null)
     {
-        return new AccessToken($response);
+        return $token;
     }
 }
--- a/src/Grant/ClientCredentials.php
+++ b/src/Grant/ClientCredentials.php
@@ -2,7 +2,7 @@
 
 namespace League\OAuth2\Client\Grant;
 
-use League\OAuth2\Client\Token\AccessToken;
+use League\OAuth2\Client\Token\AccessTokenInterface;
 
 class ClientCredentials implements GrantInterface
 {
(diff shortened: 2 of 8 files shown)


shadowhand:
I understand the intent of this change but am not sure this is best solution. A possibly better way to do this would be to create a `TokenFactory` that gets injected into each grant instance and then used like so:

```php
    public function handleResponse(array $response)
    {
        return $this->tokenFactory->create($response);
    }
```

shadowhand:
After reviewing this again, I actually really like what you are suggesting, as it centralizes the token creation with provider while allowing the grant to modify it. I would keep the code you have, with one modification: change the interface name to `TokenInterface` for easier naming compatibility.

shadowhand:
Also, this PR should be recreated against the master branch, as the 0.12.x branch is in maintenance mode.

rtheunissen:
Why can't the provider just have it's own `AccessToken`, which simply extends `League\OAuth2\Client\Token\AccessToken`?

ramsey:
Thank you so much for this, but I agree with @rtheunissen. Can providers just extend this, or is the interface needed?

ajscarvalho:
It could be done by extension, but the point remains, it must receive an Access Token object 

I just think the signature, which was
handleResponse($reponse = [])
that was changed to 
handleResponse(AccessToken $token, array $response = null)
should be just:
handleResponse(AccessToken $token)

In fact, I don't even get why you should have the handleResponse if it's allways an Identity function. Are you previewing future developments where the handleResponse method will do more that it presently does?

rtheunissen:
>  it must receive an Access Token object

Just to be clear that we're on the same page, if you extend `League\OAuth2\Client\Token\AccessToken`, that extension also qualifies as its parent type, ie. you can pass your extending class to that method as an `AccessToken`.

lightningspirit:
Yes that is true. In facto we don't need that interface. I will refactor this proposal and attach a new one without the interface and simplifying all of these.

rtheunissen:
Thanks @lightningspirit :)

ramsey:
@lightningspirit, after you refactor, make sure you run `phpcs` before pushing your changes. The Travis build is failing on this pull request due to some coding standards violations. Refer to our [contributing guide](https://github.com/thephpleague/oauth2-client/blob/master/CONTRIBUTING.md) for instructions on what to run before pushing.

Thanks!

ramsey:
@lightningspirit Any chance you can submit your changes/refactor by tomorrow? I will be cutting 1.0.0-alpha2 tomorrow (see #362).

ramsey:
I am releasing 1.0.0-beta1 on Wednesday (15 July). At that point, I am freezing the API for 1.0, so that we can move forward with a stable 1.0.0 release soon. If you'd like this to be in 1.0, please submit your changes before Wednesday. Thanks!

ramsey:
I released 1.0.0-beta1 last week, and I'm closing this since the 1.0 API is now frozen.

Please feel free to open a new pull request against the master branch, and we can work this into a future 1.1 or 1.2 release.

PR #356, opened 2015-06-26
Remove newline in header since RFC 7230 deprecates multiline headers

ramsey:
See: https://tools.ietf.org/html/rfc7230#section-3.2.4

### DIFF
--- a/src/Tool/MacAuthorizationTrait.php
+++ b/src/Tool/MacAuthorizationTrait.php
@@ -56,6 +56,6 @@ trait MacAuthorizationTrait
             $parts[] = sprintf('%s="%s"', $key, $value);
         }
 
-        return ['Authorization' => 'MAC ' . implode(",\n", $parts)];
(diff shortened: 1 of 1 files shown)


ramsey:
@shadowhand, I wanted to kick this to you for review before merging it in, since you originally wrote this. I wanted to make sure I wasn't going to break the MAC authorization header.

shadowhand:
> A field value might be preceded and/or followed by optional
   whitespace (OWS); a single SP preceding the field-value is preferred
   for consistent readability by humans.  The field value does not
   include any leading or trailing whitespace: OWS occurring before the
   first non-whitespace octet of the field value or after the last
   non-whitespace octet of the field value ought to be excluded by
   parsers when extracting the field value from a header field.

shadowhand:
Seems like it should be replaced with a space.

ramsey:
Good call. I added a space.

shadowhand:
:+1:

PR #359, opened 2015-06-29
Pass AccessToken to getDefaultHeaders

rtheunissen:
I think this was a minor thing that was missed earlier. The Reddit provider relies on token access in `getDefaultHeaders`, but I *could* restructure to go without. The current method header for `getDefaultHeaders` does expect a token though, so I'm led to believe that this is the intended behaviour.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -656,7 +656,7 @@ abstract class AbstractProvider
      */
     public function getHeaders($token = null)
     {
-        $headers = $this->getDefaultHeaders();
+        $headers = $this->getDefaultHeaders($token);
         if ($token) {
             $headers = array_merge($headers, $this->getAuthorizationHeaders($token));
         }


ramsey:
I agree that this looks like an oversight. Thanks!

shadowhand:
I don't think this is right. `getDefaultHeaders` should be usable with or without the token. Anything that requires the token should be generated in `getAuthorizationHeaders`.

rtheunissen:
I *needed* this because Reddit requires `Authorization` and `Content-Type` headers when requesting an access token. I don't want to put those in `getDefaultHeaders`, because I don't want to set them for requests that *use* the access token. I agree that the token parameter doesn't make sense here but I needed a way to determine whether I'm requesting an access token or using an access token.

Note: I've removed the parameter in #363

shadowhand:
@rtheunissen I understand the need, but the solution is wrong. The solution should be to call `getHeaders` from `getRequest` *and* make sure that `array_replace_recursive` is being used to merge existing user-defined headers.

PR #360, opened 2015-06-29
Added some AccessToken typehints where appropriate

rtheunissen:
Unless there is a good reason not to do this?

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -622,10 +622,10 @@ abstract class AbstractProvider
      *
      * Typically this is used to set Accept or Content-Type headers.
      *
-     * @param  AccessToken $token
+     * @param  AccessToken|null $token
      * @return array
      */
-    protected function getDefaultHeaders($token = null)
+    protected function getDefaultHeaders(AccessToken $token = null)
     {
         return [];
     }
@@ -638,10 +638,10 @@ abstract class AbstractProvider
      *
      * No default is provided, providers must overload this method to activate
      * authorization headers.
-     *
+     * @param AccessToken|null $token
      * @return array
      */
-    protected function getAuthorizationHeaders($token = null)
+    protected function getAuthorizationHeaders(AccessToken $token = null)
     {
         return [];
     }
@@ -651,10 +651,10 @@ abstract class AbstractProvider
      *
      * If a token is passed, the request may be authenticated through headers.
      *
-     * @param  mixed $token  object or string
+     * @param  AccessToken|null $token  object or string
      * @return array
      */
-    public function getHeaders($token = null)
+    public function getHeaders(AccessToken $token = null)
     {
         $headers = $this->getDefaultHeaders();
         if ($token) {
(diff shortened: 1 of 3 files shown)


ramsey:
I think in some cases, we were expecting that the provider might pass the string access token instead of the object to some of these methods. If we decide they must always pass an AccessToken, then I'm fine with that.

@shadowhand, any thoughts?

rtheunissen:
I believe it's in our interest to be strict with parameter types and provide flexibility with clever methods instead. So keep the rules and policies specific, but give providers enough access to adjust things accordingly. 

However, if there is a use case or requirement for passing a string, then I'm happy for us to document the expected type as `@param AccessToken|string`.

ramsey:
I agree with you. I'm just trying to make sure I'm not forgetting something. :-)

rtheunissen:
Absolutely, there's no urgency for this so I don't mind hanging around for a while.

src/Provider/AbstractProvider.php
```diff
@@ -622,10 +622,10 @@ protected function fetchUserDetails(AccessToken $token)
      *
      * Typically this is used to set Accept or Content-Type headers.
      *
-     * @param  AccessToken $token
+     * @param  AccessToken|null $token
      * @return array
      */
-    protected function getDefaultHeaders($token = null)
+    protected function getDefaultHeaders(AccessToken $token = null)
```
shadowhand: As per my comments on #359, this method shouldn't even have a `$token` parameter.

src/Tool/BearerAuthorizationTrait.php
```diff
@@ -9,7 +11,7 @@
  */
 trait BearerAuthorizationTrait
 {
-    protected function getAuthorizationHeaders($token = null)
+    protected function getAuthorizationHeaders(AccessToken $token = null)
     {
         return ['Authorization' => 'Bearer ' . $token];
```
shadowhand: This is exactly what I mean... the usage of the token here means that we explicitly **do** want to allow strings instead of instances.

src/Provider/AbstractProvider.php
```diff
@@ -638,10 +638,10 @@ protected function getDefaultHeaders($token = null)
      *
      * No default is provided, providers must overload this method to activate
      * authorization headers.
-     *
+     * @param AccessToken|null $token
      * @return array
      */
-    protected function getAuthorizationHeaders($token = null)
+    protected function getAuthorizationHeaders(AccessToken $token = null)
```
shadowhand: I don't agree with this change. One of the benefits of the `AccessToken` is that it includes a `__toString` method, which means that we don't need to care about whether we only store the token as a string, or recreate a full instance of the access token.

shadowhand:
:-1: for the reasons I listed. This offers no real benefits for end users.

src/Provider/AbstractProvider.php
rtheunissen: For instance, storing *only* the access token field of a token, then using that to create a request later?

shadowhand: Correct.

rtheunissen: Makes sense. :)

PR #361, opened 2015-06-29
Remove Zend\Diactoros dev-dependency

rtheunissen:
I don't think we need this seeing as we already have Guzzle 6 as a dependency. I know it's just a dev-dependency but it's redundant nonetheless.


ramsey:
Doh! I forgot about mocking the ResponseInterface. Thanks. :-)

rtheunissen:
:+1:

PR #363, opened 2015-06-29
Another proposed cleanup of getAccessToken

rtheunissen:
While updating the Reddit provider, I came across some cases where the current `getAccessToken` method is difficult to configure. I'll outline the changes here, but will comment inline with more detail.

- ~~Removed the request method constants.~~
- Removed `testInvalidAccessTokenMethod`
- Added `getAccessTokenUrl`
- Added `getAccessTokenHeaders`
- Added `getAccessTokenBody`
- Added `getAccessTokenOptions`
- ~~Fixed `prepareAccessTokenResponse` to also accept strings~~

- Removed `$token` parameter from `getDefaultHeaders`

These came up because:

1. I have to add a `Content-Type: application/x-www-form-urlencoded` header when requesting a Reddit access token.
2. I have to `urldecode` the request body because Reddit has a `grant_type` with forward slashes in it, which `http_build_query` automatically URL-encodes.
3. Reddit requires Basic HTTP auth when requesting an access token.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -26,12 +26,12 @@ abstract class AbstractProvider
     /**
      * @var string HTTP method used to fetch access tokens.
      */
-    const ACCESS_TOKEN_METHOD_GET = 'GET';
+    const METHOD_GET = 'GET';
 
     /**
      * @var string HTTP method used to fetch access tokens.
      */
-    const ACCESS_TOKEN_METHOD_POST = 'POST';
+    const METHOD_POST = 'POST';
 
     /**
      * @var string
@@ -355,7 +355,7 @@ abstract class AbstractProvider
      */
     protected function getAccessTokenMethod()
     {
-        return self::ACCESS_TOKEN_METHOD_POST;
+        return self::METHOD_POST;
     }
 
     /**
@@ -367,6 +367,65 @@ abstract class AbstractProvider
     protected function getAccessTokenQuery(array $params)
     {
         return http_build_query($params);
+    }
+
+    /**
+     * Returns the full URL to use when requesting an access token.
+     *
+     * @param array $params Query parameters
+     */
+    protected function getAccessTokenUrl(array $params)
+    {
+        $url = $this->getBaseAccessTokenUrl();
+
+        if ($this->getAccessTokenMethod() === self::METHOD_GET) {
+            $query = $this->getAccessTokenQuery($params);
+            return $this->appendQuery($url, $query);
+        }
+
+        return $url;
+    }
+
+    /**
+     * Returns the request body for requesting an access token.
+     *
+     * @param  array $params
+     * @return string
+     */
+    protected function getAccessTokenBody(array $params)
+    {
+        return http_build_query($params);
+    }
+
+    /**
+     * Builds request options used for requesting an access token.
+     *
+     * @param  array $params
+     * @return array
+     */
+    protected function getAccessTokenOptions(array $params)
+    {
+        $options = [];
+
+        if ($this->getAccessTokenMethod() === self::METHOD_POST) {
+            $options['body'] = $this->getAccessTokenBody($params);
+        }
+
+        return $options;
+    }
+
+    /**
+     * Returns a prepared request for requesting an access token.
+     *
+     * @param array $params Query string parameters
+     */
+    protected function getAccessTokenRequest(array $params)
+    {
+        $method  = $this->getAccessTokenMethod();
+        $url     = $this->getAccessTokenUrl($params);
+        $options = $this->getAccessTokenOptions($params);
+
+        return $this->getRequest($method, $url, null, $options);
     }
 
     /**
@@ -384,35 +443,6 @@ abstract class AbstractProvider
 
         $this->grantFactory->checkGrant($grant);
         return $grant;
-    }
-
-    /**
-     * Returns a prepared request for requesting an access token.
-     *
-     * @param array $params Query string parameters
-     */
-    protected function getAccessTokenRequest(array $params)
-    {
-        $url = $this->getBaseAccessTokenUrl();
-        $query = $this->getAccessTokenQuery($params);
-        $method = strtoupper($this->getAccessTokenMethod());
-
-        $options = [];
-
-        switch ($method) {
-            case 'GET':
-                $url = $this->appendQuery($url, $query);
-                break;
-            case 'POST':
-                $options['body'] = $query;
-                break;
-            default:
-                throw new InvalidArgumentException(
-                    "Unsupported access token request method: '$method'"
-                );
-        }
-
-        return $this->getRequest($method, $url, $options);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Provider/AbstractProvider.php
```diff
@@ -24,16 +24,6 @@
...
 
     /**
-     * @var string HTTP method used to fetch access tokens.
-     */
-    const ACCESS_TOKEN_METHOD_GET = 'GET';
-
-    /**
-     * @var string HTTP method used to fetch access tokens.
-     */
-    const ACCESS_TOKEN_METHOD_POST = 'POST';
```
shadowhand: Why are these removed? It has nothing to do with the other changes.

src/Provider/AbstractProvider.php
```diff
@@ -24,16 +24,6 @@
     const ACCESS_TOKEN_UID = null;
 
     /**
-     * @var string HTTP method used to fetch access tokens.
```
rtheunissen: I removed these because they are already constants as they stand. I like that we can accommodate any request method now, which makes these feel noisy.

shadowhand: Strings are not constants. Comparing strings to strings is messy, whereas using:

```php
if ($thing === Foo::CONST) { ... }
```

is much nicer and does not require PHP to create new temporary values in memory.

rtheunissen: HTTP request methods are constants. What if a provider uses `PUT` or some arbitrary `TOKEN` method? These should be normalised when used by the client.

src/Provider/AbstractProvider.php
```diff
@@ -370,20 +360,60 @@ protected function getAccessTokenQuery(array $params)
...
+    }
+
+    /**
+     * Returns headers that are only used when requesting an access token.
+     *
+     * @return array
+     */
+    protected function getAccessTokenHeaders()
+    {
+        return $this->getDefaultHeaders();
```
rtheunissen: This could be `[]`?

src/Provider/AbstractProvider.php
shadowhand: That doesn't make sense, because the OAuth spec clearly states that the only methods used are `GET` and `POST` as per [Section 3.1](http://tools.ietf.org/html/rfc6749#section-3.1):

> The authorization server MUST support the use of the HTTP "GET"
   method [RFC2616] for the authorization endpoint and MAY support the
   use of the "POST" method as well.

src/Provider/AbstractProvider.php
```diff
@@ -370,20 +360,60 @@ protected function getAccessTokenQuery(array $params)
...
+
+    /**
+     * Returns the request body for requesting an access token.
+     *
+     * @param  array $params
+     * @return string
+     */
+    protected function getAccessTokenBody(array $params)
+    {
+        return http_build_query($params);
```
rtheunissen: This is a distinction from `getAccessTokenQuery` because the URL and POST body may not always have the same parameters and encoding.

src/Provider/AbstractProvider.php
```diff
@@ -370,20 +360,60 @@ protected function getAccessTokenQuery(array $params)
...
+    }
+
+    /**
+     * Builds request options used for requesting an access token.
+     *
+     * @param  array $params
+     * @return array
+     */
+    protected function getAccessTokenRequestOptions(array $params)
```
rtheunissen: This feels a bit Guzzle-specific but fits nicely with `getRequest`.

shadowhand:
I'm fairly sure that: a) these changes are not necessary, b) the functionality for sending form data already exists (but may be partially broken).

src/Provider/AbstractProvider.php
```diff
@@ -581,12 +612,12 @@ protected function parseResponse(ResponseInterface $response)
      * Custom mapping of expirations, etc should be done here. Always call the
      * parent method when overloading this method!
      *
-     * @param  array $result
+     * @param  mixed $result
```
rtheunissen: `parseResponse` can return a string, *but* if a uid key is to be set and the parse didn't result in an array, that means the parse is broken. Not sure how to indicate that.

src/Provider/AbstractProvider.php
rtheunissen: I'll restore them, you make a good point. :+1:

src/Provider/AbstractProvider.php
shadowhand: `parseResponse` should handle that.

rtheunissen: Yup, agreed.

src/Provider/AbstractProvider.php
rtheunissen: Restored. :+1:

rtheunissen:
@shadowhand 
> these changes are not necessary

Sure, they are not *necessary*. Doesn't mean they're not an improvement. If this was already a stable release out in the wild I wouldn't bother. I'd find a way to make my own extension work with it. But we're trying to craft this as well as we can while it's new. These changes a) allows for a much cleaner Reddit provider (and I'm sure others too), maintains 100% coverage, and doesn't *reduce* functionality in any way. 

> the functionality for sending form data already exists

Yup, and now it still exists, albeit in a slightly different form. Before we were setting the `body` option as the same value as `$query`, assuming the same encoding. Why not split them up into build body and build query? An alternative could be to go with something generic like `getAccessTokenData`.

src/Provider/AbstractProvider.php
```diff
@@ -355,7 +355,7 @@ protected function appendQuery($url, $query)
      */
     protected function getAccessTokenMethod()
     {
-        return self::ACCESS_TOKEN_METHOD_POST;
+        return 'POST';
```
shadowhand: This still needs to be changed.

test/src/Provider/AbstractProviderTest.php
```diff
@@ -472,22 +472,7 @@ public function testGetAccessToken($method)
         $this->assertSame($raw_response['access_token'], $token->getToken());
         $this->assertSame($raw_response['expires'], $token->getExpires());
     }
-
-    /**
-     * @expectedException InvalidArgumentException
-     */
-    public function testInvalidAccessTokenMethod()
```
shadowhand: lost this check

rtheunissen: Is it necessary?

src/Provider/AbstractProvider.php
rtheunissen: Fixed in 88f41007c29febf44570dcd17c6f9eab3ab2400c

src/Provider/AbstractProvider.php
rtheunissen: Fixed in 7ac78334964f1aca34113f773f54135eecc56b47

src/Provider/AbstractProvider.php
rtheunissen: Potential issue is that `getDefaultHeaders` is only called via `getHeaders` when using an access token, *not* when requesting an access token.

shadowhand: I think this would be better solved by having `getRequest` also call `getHeaders`. Also, I noticed another thing... any time `getHeaders` is called by `get*Request`, it should be using `array_replace` instead of directly setting `$options['headers']` otherwise user-defined headers will be lost:

```php
$options = array_replace_recursive($options, [
    'headers' => $this->getHeaders(...),
]);
```

rtheunissen: That makes sense, in both regards.

PR #364, opened 2015-06-29
Proposed refactor of getRequest and getAccessToken

rtheunissen:
This is a cleaner attempt at #363.

I'll outline the changes here and comment inline with more details.

---

- **Renamed** `ACCESS_TOKEN_METHOD_GET` -> `METHOD_GET`
- **Renamed** `ACCESS_TOKEN_METHOD_POST` -> `METHOD_POST`
- ~~**Removed** `getAuthenticatedRequest`~~ 
- **Removed** `$token` parameter from `getDefaultHeaders`
- **Removed** `testInvalidAccessTokenMethod`
- **Added** `createRequest`
- **Added** `getAccessTokenUrl`
- **Added** `getAccessTokenBody`
- **Added** `getAccessTokenOptions`
- ~~**Added** `$token` parameter to `getRequest`~~

---

These came up because you have to add a `Content-Type: application/x-www-form-urlencoded` header when requesting a Reddit access token.

You have to `urldecode` the request body because Reddit has a `grant_type` with forward slashes in it, which `http_build_query` automatically encodes and Reddit doesn't like it. I'm not sure if setting the urlencoded header fixes that.

Reddit requires Basic HTTP auth when requesting an access token, so should have access to headers.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -442,7 +442,7 @@ abstract class AbstractProvider
         $url     = $this->getAccessTokenUrl($params);
         $options = $this->getAccessTokenOptions($params);
 
-        return $this->getRequest($method, $url, null, $options);
+        return $this->getRequest($method, $url, $options);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Provider/AbstractProvider.php
```diff
@@ -387,32 +387,62 @@ protected function verifyGrant($grant)
...
+        return $url;
+    }
+
+    /**
+     * Returns the request body for requesting an access token.
+     *
+     * @param  array $params
+     * @return string
+     */
+    protected function getAccessTokenBody(array $params)
```
rtheunissen: This is *almost* identical to `getAccessTokenQuery`. I would be okay with combining them, but I don't want to use `$options['body'] = $query`.

src/Provider/AbstractProvider.php
```diff
@@ -442,36 +471,25 @@ public function getAccessToken($grant, array $options = [])
...
-     * The request is not automatically authenticated.
      *
      * @param  string $method
      * @param  string $url
+     * @param  AccessToken|string $token
      * @param  array  $options Any of "headers", "body", and "protocolVersion".
      * @return RequestInterface
      */
-    public function getRequest($method, $url, array $options = [])
+    public function getRequest($method, $url, $token = null, array $options = [])
```
shadowhand: I don't really like this. In my opinion having separate public methods for authenticated and non-authenticated requests is valuable. Having to pass a `null` in the middle of a method call is a good sign that separate methods are useful, and having a hard `$token` requirement for authenticated requests allows for more accurate hinting.

src/Provider/AbstractProvider.php
```diff
@@ -387,32 +387,62 @@ protected function verifyGrant($grant)
...
-            case 'POST':
-                $options['body'] = $query;
-                break;
-            default:
-                throw new InvalidArgumentException(
-                    "Unsupported access token request method: '$method'"
-                );
+        if ($this->getAccessTokenMethod() === self::METHOD_POST) {
+            $options['body'] = $this->getAccessTokenBody($params);
         }
```
shadowhand: I still feel that we are losing something by separating GET/POST request building. We're definitely losing method validation, which should be enforced to meet the OAuth 2 RFC.

I guess I'd like to see a much smaller change to allow for the modifications you want, maybe just a one line change?

```php
$options = $this->getAccessTokenOptions($options, $params);
```

Which would be called just before `getRequest` to allow the provider to modify the options without overloading anything else?

rtheunissen: > We're definitely losing method validation, which should be enforced to meet the OAuth 2 RFC

Isn't that way out of scope of this library? Meeting the OAuth2 spec is up to the auth provider (ie. Google, Facebook etc.). We should be able to assume that the spec has been met. In effect, we're checking of the auth provider implemented OAuth2 correction -- not at all the place for that.

src/Provider/AbstractProvider.php
rtheunissen: I thought about this pretty hard before I went ahead with it. When would a provider (or API wrapper) want to make an request that is not authenticated? The only time that that make sense is for the request to authenticate, which is internal.

> having a hard $token requirement for authenticated requests allows for more accurate hinting

Didn't you argue against strong access token hinting earlier? Should be able to get a request using a string-value token, so I don't think accurate hinting is an argument here.

rtheunissen: Having thought about this for a while, I agree with you. 

See bbfe83e3d01f8af39e1d7f379b4d0ace2f55de3d

src/Provider/AbstractProvider.php
ramsey: I like keeping them separate, too.

src/Provider/AbstractProvider.php
ramsey: I agree with @rtheunissen here. While my personal preference is to strictly adhere to the spec, we're dealing with providers who don't, so we have to be flexible enough to accommodate them.

PR #365, opened 2015-06-30
Remove --prefer-source from .travis.yml

rtheunissen:
> **--prefer-source**: There are two ways of downloading a package: `source` and `dist`. For stable versions Composer will use the `dist` by default. The `source` is a version control repository. If `--prefer-source` is enabled, Composer will install from `source` if there is one. This is useful if you want to make a bugfix to a project and get a local git clone of the dependency directly.

Travis CI clones the repository of every dependency every time. I think this is why our builds are timing out, and there is no reason why we need to be doing that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ php:
 
 before_script:
   - travis_retry composer self-update
-  - travis_retry composer install --no-interaction --prefer-source
+  - travis_retry composer install --no-interaction
 
 script:
   - mkdir -p build/logs


rtheunissen:
Nope. 
screen shot 2015-06-30 at 2 26 07 pm

shadowhand:
When you use dist versions, it pings Github to get the tag list, which will cause the (anonymous) API rate limit to be reached fairly quickly with popular projects.

rtheunissen:
Oh cool, didn't know that. Thanks.

PR #371, opened 2015-07-03
Dealing with multiple access token URLs

### LINKED ISSUE #368: Deal with different base access token url between authorization and refresh
I encounter a provider that has different base access token url between authorization and refresh. Current v1 abstract provider class doesn't take different grant type into account when getting base access token url. How to deal with this situation properly?

rtheunissen:
Attempts to fix #368

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -214,7 +214,7 @@ abstract class AbstractProvider
     }
 
     abstract public function getBaseAuthorizationUrl();
-    abstract public function getBaseAccessTokenUrl();
+    abstract public function getBaseAccessTokenUrl(array $params);
     abstract public function getUserDetailsUrl(AccessToken $token);
 
     /**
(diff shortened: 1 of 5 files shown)


shadowhand:
I think this shows that there is a flaw in the abstraction of `getAccessTokenRequest`. The user needs _all_ of the available input when deciding how to build out the request, including `$grant` and (possibly) `$options`.

rtheunissen:
@shadowhand I disagree, the only flaw is not passing enough information to `getAccessTokenUrl`. . We were once happy to not pass *anything* to `getAccessTokenUrl`, in fact you yourself implemented it as a constant. We reverted back to abstract method which a provider has to implement, giving us some more flexibility. This case shows brought to our attention that some providers have varying access token URL's -- something none of us had thought of. 

Also note that this PR is already reflected in #364 [here](https://github.com/rtheunissen/oauth2-client/blob/access_token_cleanup/src/Provider/AbstractProvider.php#L442)

rtheunissen:
Fixing the tests now.

shadowhand:
Ah yes I see now. I made an invalid assumption about how the `refresh_token` parameter was passed around. Your solution should work correctly.

rtheunissen:
:+1:

ramsey:
Looks good. :+1:

PR #375, opened 2015-07-04
Enhanced json_encode() serialization of AccessToken

### LINKED ISSUE #373: accessToken VS access_token
Hi,

The AccessToken.php file changes the tokens key names.
From "access_token" to "accessToken" for example.

As Google expects me to use access_token, it fails if I used my token array without changing the keys first.

Is it wanted for some reasons?

ramsey:
Fixes #373

@shadowhand and @rtheunissen, please review and let me know if you think this is a good idea. I'm not sure I'm sold on it, but it definitely adds convenience for those using `json_encode()` on the `AccessToken` object.

### DIFF
--- a/src/Token/AccessToken.php
+++ b/src/Token/AccessToken.php
@@ -15,9 +15,10 @@
 namespace League\OAuth2\Client\Token;
 
 use InvalidArgumentException;
+use JsonSerializable;
 use RuntimeException;
 
-class AccessToken
+class AccessToken implements JsonSerializable
 {
     /**
      * @var  string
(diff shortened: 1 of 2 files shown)


rtheunissen:
I think it's fine because we know there's a use case for it. However there might be a discrepancy between expire time and time until expires, because providers seem to produce those differently. 

Other than that, looks good.

koxon:
+1 for this

ramsey:
The spec uses `expires_in`, which is why I'm converting it to that.

rtheunissen:
:+1:

shadowhand:
Awesome!

Daanvm:
@ramsey Wouldn't it be better to just save `expires` instead of `expires_in`? When I save a json serialized version of an access token and restore it later, the `expires_in` value isn't correct anymore, but `expires` is. Am I missing something?

ramsey:
We're following the RFC, which uses `expires_in`. That said, I'll give this some thought, since you're correct that `expires_in` has little meaning beyond the initial response from the service.

Daanvm:
I can create a pull request if you want to? Just wanted to ask you before I'd make any changes.

ramsey:
My primary concern is in changing/removing an implementation detail that others might rely on. However,  since the constructor for AccessToken can handle both `expires` and `expires_in` passed to it as options, then maybe it's not a big deal?

@shadowhand, thoughts?

shadowhand:
I think in this scenario, we need to go for correctness, which can only be done by using `expires`. The spec is using `expires_in` because it is dealing with an **active** request/response lifecycle, whereas this is about storage.

ramsey:
Yeah. It was a bad decision on my part to use `expires_in`. We can create a version 1.1 for this and note it as a potentially breaking change in the changelog.

PR #376, opened 2015-07-07
Convert all uses of "user" to "resource owner"

ramsey:
This is an exploratory change to get feedback from the oauth2-client community. It is the result of a conversation on Twitter with @neilcrookes: https://twitter.com/ramsey/status/618437720554143744

I have never been comfortable providing any entity that represents a user, since the OAuth 2.0 specification does not specify anything to do with users. However, the specification does talk about resource owners, which may be end-users, if the resource owner is a person (see [section 1.1 of the OAuth 2.0 spec](http://tools.ietf.org/html/rfc6749#section-1.1)), but it may not be a user at all, as is the case with @neilcrookes's use-case.

With this change, the concept of the "user" is migrated to that of a "resource owner." Individual providers may call their class implementing the new `ResourceOwnerInterface` a User, if the resource owner in their service is a person, or they may call it something else entirely.

Here is exactly what has changed:

Used to be | Changed to
---------- | ----------
`AbstractProvider::ACCESS_TOKEN_UID` | `AbstractProvider::ACCESS_TOKEN_RESOURCE_OWNER_ID`
`AbstractProvider::getUserDetailsUrl()` | `AbstractProvider::getResourceOwnerDetailsUrl()`
`AbstractProvider::createUser()` | `AbstractProvider::createResourceOwner()`
`AbstractProvider::getUser()` | `AbstractProvider::getResourceOwner()`
`AbstractProvider::fetchUserDetails()` | `AbstractProvider::fetchResourceOwnerDetails()`
`UserInterface` | `ResourceOwnerInterface`
`UserInterface::getUserId()` | `ResourceOwnerInterface::getId()`
`StandardUser` | `StandardResourceOwner`
`AccessToken::getUid()` | `AccessToken::getResourceOwnerId()`

To summarize, I changed "User" to "ResourceOwner", "UID" to "RESOURCE_OWNER_ID", "Uid" to "ResourceOwnerId", and "uid" to "resource_owner_id". I have updated all the tests, and everything passes, with 100% code coverage.

Additionally, I added the `AbstractProvider::getAccessTokenResourceOwnerId()` method. I noticed that this was being called in the `StandardProvider`, though the method was not defined anywhere.

### DIFF
--- a/README.PROVIDER-GUIDE.md
+++ b/README.PROVIDER-GUIDE.md
@@ -22,45 +22,48 @@ and implement the required abstract methods:
 ```php
 abstract public function getBaseAuthorizationUrl();
 abstract public function getBaseAccessTokenUrl(array $params);
-abstract public function getUserDetailsUrl(AccessToken $token);
+abstract public function getResourceOwnerDetailsUrl(AccessToken $token);
 abstract protected function getDefaultScopes();
 abstract protected function checkResponse(ResponseInterface $response, $data);
-abstract protected function createUser(array $response, AccessToken $token);
+abstract protected function createResourceOwner(array $response, AccessToken $token);
 ```
 
 Each of these abstract methods contain a docblock defining their expectations
 and typical behavior. Once you have extended this class, you can simply follow
 the [usage example in the README](README.md#usage) using your new `Provider`.
 
-### Account identifiers in access token responses
+### Resource owner identifiers in access token responses
 
-We have decided to abstract away as much of the user details as possible, since
-these are not part of the OAuth 2.0 specification and are very specific to each
+In services where the resource owner is a person, the resource owner is sometimes
+referred to as an end-user.
+
+We have decided to abstract away as much of the resource owner details as possible,
+since these are not part of the OAuth 2.0 specification and are very specific to each
 service provider. This provides greater flexibility to each provider, allowing
-them to handle the implementation details for service users.
+them to handle the implementation details for resource owners.
 
-The `AbstractProvider` does not specify an access token user identifier. It is
-the responsibility of the provider class to set the `ACCESS_TOKEN_UID` constant
+The `AbstractProvider` does not specify an access token resource owner identifier. It is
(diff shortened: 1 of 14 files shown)


ramsey:
Requesting feedback and review from oauth2-client providers, especially: @shadowhand @rtheunissen @SammyK @stevenmaguire

rtheunissen:
Makes sense to me. For the sake of brevity, could change `ResourceOwner` to `Owner`? Looks good though, :+1:

ramsey:
I like brevity, too, but I think clarity is more important. Since the spec calls these entities "resource owners," I chose to use `ResourceOwner` in referring to them in code.

rtheunissen:
> Since the spec calls these entities "resource owners," I chose to use ResourceOwner

:dart:

shadowhand:
Hmm, okay. I feel pretty ambiguous about this change, but being consistent with spec is good.

SammyK:
:+1: For making it closer to DSL of OAuth spec! :)

> I like brevity, too, but I think clarity is more important.

I'm also a big fan of this philosophy as well. Admittedly it look me a while to figure out what the "o" was in `AbstractProvider::ACCESS_TOKEN_OID` and `AccessToken::getOid()`. I'd lean to being more explicit: `AbstractProvider::ACCESS_TOKEN_RESOURCE_OWNER_ID` and `AccessToken::getResourceOwnerId()`.

But even without those changes this is a good idea. :)

rtheunissen:
:+1: @SammyK

ramsey:
Merging this in, which will probably cause #378 to be in conflict.

shadowhand:
No problem, I'll rebase.

ramsey:
Oh. First, I'll add @SammyK's recommendations. Then merge.

stevenmaguire:
I'm a :+1: as well. I like moving towards intuitive design based on a spec. I just got 6 provider packages upgraded to 1.0 alpha, so they are "on the bench" and ready for more changes. Great work!

PR #380, opened 2015-07-11
Storing the Provider Response in a class attribute

ssnukala:
Would like to propose storing the response in a class attribute, so
that if we configure the OAuth to get additional information than the
User Object has, then this attribute can be used to get that
information. For instance I was trying to get “summary” and “Positions”
fields from LinkedIn, I am able to set $fields to include that in the
call to linked in but the return object only returns the User object
and that does not have all the fields. So providing a way to access the
actual response received would be good so that users can use this
extended information.

I am using something like this

        $oaFields = ['id', 'email-address', 'first-name', 'last-name', 'headline',
        'location', 'industry', 'picture-url', 'public-profile-url','summary','specialties','positions'];

        $this->_provider = new \League\OAuth2\Client\Provider\LinkedIn([
            'clientId' => $clientId,
            'clientSecret' => $clientSecret,
            'redirectUri' => $this->_app->site->uri['public'] . "/oauth/linkedin/$callback_page",
            'scopes' => $scopes]);
        $this->_provider->fields=$oaFields;

So even though the fetchUserDetails gets all the data from LinkedIn the User object does not pass all this back.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -35,6 +35,8 @@ abstract class AbstractProvider implements ProviderInterface
 
     public $authorizationHeader;
 
+    public $providerResponse;
+    
     /**
      * @var GuzzleClient
      */
--- a/src/Provider/LinkedIn.php
+++ b/src/Provider/LinkedIn.php
@@ -34,6 +34,8 @@ class LinkedIn extends AbstractProvider
 
     public function userDetails($response, AccessToken $token)
     {
+        $this->providerResponse=$response;
+
         $user = new User();
 
         $email = (isset($response->emailAddress)) ? $response->emailAddress : null;


rtheunissen:
I think this comes down to the LinkedIn provider's handling of those extra fields. Opening an issue on [the LinkedIn provider's repo](https://github.com/thephpleague/oauth2-linkedin) might be the best way to find a solution and make it happen.

ssnukala:
I guess part of it is on the provider but it would be good to expose the complete response rather than limiting it to the user object. The data is there but is not available. So if we do it in the abstract class then every providers response data can be exposed.

shadowhand:
That wouldn't provide a contract (interface) of what data is available. Storing a blob of data is never as ideal as exposing specific methods to access data. Hence, LinkedIn provider should be using a custom ResourceOwner (formerly: User) class that exposes the proper data.

ramsey:
Thank you for your interest and time in contributing this pull request.

I agree with @rtheunissen and @shadowhand. Please take a look at 1.0.0-alpha2 and the [LinkedIn provider](https://github.com/thephpleague/oauth2-linkedin), as @rtheunissen recommended, which is based on 1.0 of this library. The individual providers are where this type of information should be exposed.

PR #381, opened 2015-07-14
adding content-type header to request to retrieve access token

stevenmaguire:
Referenced in #348

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -445,7 +445,7 @@ abstract class AbstractProvider
      */
     protected function getAccessTokenOptions(array $params)
     {
-        $options = [];
+        $options = ['headers' => ['content-type' => 'application/x-www-form-urlencoded']];
 
         if ($this->getAccessTokenMethod() === self::METHOD_POST) {
             $options['body'] = $this->getAccessTokenBody($params);


src/Provider/AbstractProvider.php
```diff
@@ -465,7 +465,8 @@ protected function getAccessTokenRequest(array $params)
         $url     = $this->getAccessTokenUrl($params);
         $options = $this->getAccessTokenOptions($params);
 
-        return $this->getRequest($method, $url, $options);
+        return $this->getRequest($method, $url, $options)
+            ->withHeader('content-type', 'application/x-www-form-urlencoded');
```
shadowhand: This isn't quite right... You want to define `$options['headers']` in `getAccessTokenOptions`, right after the `$options['body']` define on line 451.

stevenmaguire:
Ah, I missed that options method! Much better!

rtheunissen:
@stevenmaguire take a look at #382 and #383, see if either of the solves this problem, or which one you think makes more sense?

stevenmaguire:
Based on the feedback @shadowhand provided on the two alternatives, I've amended the proposed solution here, to set the headers to the specific content-type header needed for the request. My feeling is that if specific providers need to include additional configuration they could override this method as needed, and the basic functionality still exists.

Thoughts?

rtheunissen:
My only concern with this approach is that it feels hard-coded, regardless of the request method used by the provider. I don't like that you'd have to override `getAccessTokenOptions` to adjust only the headers. This isn't *bad* though, I'm just not convinced that it's the best solution.

Any reason why we can't use `getDefaultHeaders`? See #383

src/Provider/AbstractProvider.php
```diff
@@ -449,6 +449,7 @@ protected function getAccessTokenOptions(array $params)
 
         if ($this->getAccessTokenMethod() === self::METHOD_POST) {
             $options['body'] = $this->getAccessTokenBody($params);
+            $options['headers'] = ['content-type', 'application/x-www-form-urlencoded'];
```
shadowhand: Upon reading [the spec](http://tools.ietf.org/html/rfc6749#section-4.1.3) again it is clear that this is correct content type for _both_ GET and POST requests. This is the correct solution, it just needs to be outside of the `if` statement so that it applies to any request.

shadowhand:
@rtheunissen 
> Any reason why we can't use `getDefaultHeaders`?

src/Provider/AbstractProvider.php
rtheunissen: :+1:

rtheunissen:
> Most requests to the authorized endpoints will (likely) be application/json

I highly doubt that? Only the odd PATCH might be?

shadowhand:
@rtheunissen not to the OAuth endpoints, but to the API endpoints. Remember that `getDefaultHeaders` is used by the generic `getRequest` and `getAuthorizedRequest` methods.

rtheunissen:
@shadowhand Yeah, to the API endpoints. It's very rare to have the request body be JSON. I'm aware that `getDefaultHeaders` is used by the generic `getRequest`, but also that any headers set in the request options or authorized headers would replace the ones in `getDefaultHeaders`.

shadowhand:
In my experience, if an API is using JSON outputs, they also expect JSON input as POST and PUT. As examples, most of the OAuth clients I've written (Square, Clover, Twitter, etc) all use JSON bodies.

rtheunissen:
:+1:

stevenmaguire:
Great catch @shadowhand. Solution updated!

src/Provider/AbstractProvider.php
```diff
@@ -446,6 +446,7 @@ protected function getAccessTokenBody(array $params)
     protected function getAccessTokenOptions(array $params)
     {
         $options = [];
+        $options['headers'] = ['content-type', 'application/x-www-form-urlencoded'];
```
shadowhand: could be a single line, but no harm.

stevenmaguire: I like one line better! Sometimes "coding style" opinions show up ;) 

If you are in support of one line here, so am I. Commit updated.

shadowhand:
Just as a side note, I really like to see commits follow this form:

> [When applied, this commit will ...] commit message

It makes the tense "active" and provides a expectation. In this form, your commit message would be written:

> Add content-type header to access token request

Fairly minor difference, but probably easier to read and understand when reading log history.

:shipit:

stevenmaguire:
:+1: good feedback!

ramsey:
Thanks everyone for reviewing and providing feedback!

mstephens:
Hi guys

As per @shadowhand comment, is there anyway to have the ability to override the content type on the `getAccessTokenOptions` method? 

We have come across quite a few OAuth providers that required in `application/json` rather than `application/x-www-form-urlencoded` and feel this shouldn't be hard-coded within the logic.

Great repo by the way!

Cheers
Matt

shadowhand:
@mstephens any provider can overload the [`getAccessTokenBody()`](https://github.com/thephpleague/oauth2-client/blob/a1ad0d014dbefd17fe900c88469e917d9e16a81f/src/Provider/AbstractProvider.php#L509-L512) or `getAccessTokenOptions()` methods to achieve the correct type of body required for the provider.

The number of providers that do not use/allow form encoding is fairly rare in my experience.

PR #382, opened 2015-07-14
Alternative solution to set access token request headers

rtheunissen:
This PR is an alternative to #381, aiming to solve #340 and #348.

Makes use of a new method called `getAccessTokenRequestHeaders` and a trait called `UrlEncodedRequestTrait`.

The only thing I'm thinking here is why not use `getDefaultHeaders` instead? See #383

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -445,7 +445,9 @@ abstract class AbstractProvider
      */
     protected function getAccessTokenOptions(array $params)
     {
-        $options = [];
+        $options = [
+          'headers' => $this->getAccessTokenRequestHeaders($params),
+        ];
 
         if ($this->getAccessTokenMethod() === self::METHOD_POST) {
             $options['body'] = $this->getAccessTokenBody($params);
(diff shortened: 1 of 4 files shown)


src/Provider/AbstractProvider.php
```diff
@@ -445,7 +445,9 @@ protected function getAccessTokenBody(array $params)
      */
     protected function getAccessTokenOptions(array $params)
     {
-        $options = [];
+        $options = [
+          'headers' => $this->getAccessTokenRequestHeaders($params),
+        ];
 
         if ($this->getAccessTokenMethod() === self::METHOD_POST) {
             $options['body'] = $this->getAccessTokenBody($params);
```
shadowhand: This is the only condition that requires having headers. Sending `application/x-www-form-urlencoded` for a `METHOD_GET` would be wrong.

src/Provider/AbstractProvider.php
```diff
@@ -703,6 +705,17 @@ protected function fetchResourceOwnerDetails(AccessToken $token)
     }
 
     /**
+     * Returns the headers used when requesting an access token.
+     *
+     * @param  array $params
+     * @return array
+     */
+    protected function getAccessTokenRequestHeaders(array $params)
```
shadowhand: I don't feel like this is necessary.

src/Provider/AbstractProvider.php
rtheunissen: You wouldn't use the trait for your provider if you're using GET. It'll just be the default empty headers.

PR #383, opened 2015-07-14
Second alternative solution to set access token request headers

rtheunissen:
This PR is an alternative to #382 , aiming to solve #340 and #348.

Uses `getDefaultHeaders` rather than creating a new `getAccessTokenRequestHeaders` method.


shadowhand:
This doesn't make sense to me.

rtheunissen:
@shadowhand Could you please explain why? Saying it doesn't make sense doesn't add much value to the discussion.

PR #387, opened 2015-07-20
Add support for nested index within Access Token Resource Owner ID

stevenmaguire:
The Instagram API returns a payload with a nested `User` object that contains the `uid` of  the resource owner.

```json
{
    "access_token": "fb2e77d.47a0479900504cb3ab4a1f626d174d2d",
    "user": {
        "id": "1574083",
        "username": "snoopdogg",
        "full_name": "Snoop Dogg",
        "profile_picture": "..."
    }
}
```

Currently, the providers only support "root" indexes. This solution provides support to access nested indexes using dot notation.

This solution would allow provider implementations to indicate nested resource owner id's in access token responses.

```php
    /**
     * @var string Key used in the access token response to identify the resource owner.
     */
    const ACCESS_TOKEN_RESOURCE_OWNER_ID = 'user.id';
```

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -657,7 +657,10 @@ abstract class AbstractProvider
     protected function prepareAccessTokenResponse(array $result)
     {
         if ($this->getAccessTokenResourceOwnerId()) {
-            $result['resource_owner_id'] = $result[$this->getAccessTokenResourceOwnerId()];
+            $result['resource_owner_id'] = $this->getValueByKey(
+                $this->getAccessTokenResourceOwnerId(),
+                $result
+            );
         }
         return $result;
     }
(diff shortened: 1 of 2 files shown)


ramsey:
Looks good to me. Thanks!

shadowhand:
I'm not entirely sure I agree with these changes being done in the root project. The LinkedIn provider could simply overload `prepareAccessTokenResponse` to achieve the same result.

@ramsey doesn't this modify the API, therefore making it an invalid change within the 1.0 release?

ramsey:
@shadowhand It doesn't change any classes, constants, methods, or method signatures, so I don't consider it a change to the API.

You could certainly overload `prepareAccessTokenResponse`, but I thought this would provide a good convenient way of handling this case for all projects and especially for the GenericProvider case.

stevenmaguire:
:+1:

PR #389, opened 2015-07-25
Update README.md to reflect changes in Token class

auburus:
Just noticed this, it's my first pull request (like... ever) so if anything more should be done, please just tell me.


ramsey:
Thanks and congrats on your first pull request!

PR #392, opened 2015-07-31
Always attempt to parse as JSON and fallback on failure

ramsey:
I've run into a few cases where the content-type returned is `text/html` while the body is JSON. In this way, we attempt to parse it as JSON, and if it fails, we fall back to just returning the content as is.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -641,7 +641,13 @@ abstract class AbstractProvider
             return $parsed;
         }
 
-        return $content;
+        // Attempt to parse the string as JSON anyway,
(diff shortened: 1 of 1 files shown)


ramsey:
Looking for feedback on this, @shadowhand, @rtheunissen.

shadowhand:
I don't love it, but it is practical.

ramsey:
I don't love it, either, which is mostly why I was asking for other eyes to take a look. :-)

rtheunissen:
You could probably ditch the 'json' check then if you're going to be trying it anyway. I don't like depending on the exception here but like @shadowhand said, it's practical.

PR #393, opened 2015-07-31
Use the Bearer authorization header for the generic provider

ramsey:
The `GenericProvider` was failing to send authenticated requests to anything, since it didn't specify any authorization headers (or provide a means to set them). I have opted to default to the Bearer token authorization header for the generic provider.

### DIFF
--- a/src/Provider/GenericProvider.php
+++ b/src/Provider/GenericProvider.php
@@ -17,10 +17,13 @@ namespace League\OAuth2\Client\Provider;
 use InvalidArgumentException;
 use League\OAuth2\Client\Provider\Exception\IdentityProviderException;
 use League\OAuth2\Client\Token\AccessToken;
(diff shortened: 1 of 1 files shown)


src/Provider/GenericProvider.php
```diff
@@ -172,6 +172,11 @@ protected function getScopeSeparator()
         return $this->scopeSeparator ?: parent::getScopeSeparator();
     }
 
+    protected function getAuthorizationHeaders($token = null)
+    {
+        return ['Authorization' => 'Bearer ' . $token];
```
shadowhand: I think there is a trait for this.

ramsey:
Another one where I'd like some feedback please, in case you think we should do this differently. @shadowhand, @rtheunissen

ramsey:
Rebased and fixed up the commit to use the trait.

shadowhand:
Hooray! :shi

PR #395, opened 2015-08-04
[WIP] Working examples to illustrate use of the oauth2-client library

ramsey:
This pull request is a work-in-progress.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,15 @@
+# league/oauth2-client Examples
+
+These examples require PHP 5.5 or greater to run. To run the examples, use
+PHP's built-in web server:
+
+```
+cd examples/
+php -S localhost:8000
+```
+
+In a web browser, go to <http://localhost:8000>.
+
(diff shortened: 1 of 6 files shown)


shadowhand:
FWIW I have a demo repository up and running at [oauth2-demos][/shadowhand/oauth2-demos]). I'll try this out right now.

ramsey:
Doh! I didn't mean to duplicate work. We don't need to include these.

shadowhand:
@ramsey the content you have here is really fantastic though! If you can make a PR to merge what you've written up with my implementation, it would be a stellar combination of documentation and demonstration.

shadowhand:
Also I am very much using a framework and the code might not be super easy to follow. I'm not really inclined to drop the framework, but perhaps explaining the calls like we do in the README is helpful?

ramsey:
Closing this pull request. I'll either redirect it to fit into oauth2-demos or figure out a better place for the content to live.

PR #396, opened 2015-08-06
Add array typehint to provider constructor

rtheunissen:


shadowhand:
The only downside to this is that it prevents using `ArrayObject` or any `Iterable` from being used as the `$options`. Not saying we shouldn't do it, just something to consider.

rtheunissen:
@shadowhand that would be a case for all `array` typehints. :( 

I'd rather keep consistent with the documentation.

shadowhand:
:+1:

PR #397, opened 2015-08-06
Skip JSON content type check

rtheunissen:
Seeing as we're attempting to parse JSON anyway, I don't see the point of checking the JSON content type. I don't really like how we're using the exception here but it makes sense from a practical point of view.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -661,10 +661,15 @@ abstract class AbstractProvider
         }
 
         // Attempt to parse the string as JSON regardless of content type,
-        // since some providers use non-standard content types.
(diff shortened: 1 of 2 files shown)


shadowhand:
Makes sense to me.

ramsey:
Looks good, but there are test failures, so please check on that.

rtheunissen:
I missed that an `UnexpectedValueException` should be thrown if a JSON content type was set but the JSON parse failed, but *not* be thrown when a JSON content type was not set.

It's very similar to master, the only difference is that the JSON content type check is only done if the parse actually failed. So the gain isn't significant, but still of value.

Also refactored the tests for `parseContent` a bit. :hammer:

PR #398, opened 2015-08-06
Add toArray to ResourceOwnerInterface

shadowhand:
Provides a more consistent way to fetch whatever information is available about the resource owner, without assuming any particular structure.

### DIFF
--- a/src/Provider/ResourceOwnerInterface.php
+++ b/src/Provider/ResourceOwnerInterface.php
@@ -25,4 +25,11 @@ interface ResourceOwnerInterface
      * @return mixed
      */
     public function getId();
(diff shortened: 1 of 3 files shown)


shadowhand:
I don't think this is unreasonable and it helps make the API consistent with other providers.

ramsey:
I agree. This would help a dev in a recent conversation on an oauth2-facebook issue.

stevenmaguire:
@ramsey since this change has updated the API, do you plan to cut a new beta release?

ramsey:
Good question. I'm traveling today. I'll get a new beta release out in the morning.

PR #400, opened 2015-08-12
Add class-level docblocks and clean up some documentation

ramsey:


src/Provider/AbstractProvider.php
```diff
@@ -126,7 +138,7 @@ public function __construct(array $options = [], array $collaborators = [])
      * Sets the grant factory instance.
      *
      * @param  GrantFactory $factory
-     * @return $this
+     * @return self
```
shadowhand: Is this the standard way to hint this? I've been using `$this` for years... doing it wrong?

ramsey: I'm not sure what the standard is, but if you use `$this`, the generated documentation just contains the string "$this." If you use `self`, the generated documentation provides a link to the current class.

Of course, this could differ from one implementation to another, because there is no standard. I've been using ApiGen to generate the docs. :-)

src/Token/AccessToken.php
```diff
@@ -41,7 +46,11 @@ class AccessToken implements JsonSerializable
     protected $resourceOwnerId;
 
     /**
-     * @param array $options
+     * Constructs an access token.
+     *
+     * @param array $options An array of options returned by the service provider
+     *     in the access token request. The `access_token` option is required.
```
shadowhand: I'm really unsure about adding this much documentation to `@param` hints. The code should be the definitive source of what is happening and these kinds of things tend to get out of date and end up causing more harm than good.

src/Grant/AbstractGrant.php
```diff
@@ -16,6 +16,18 @@
...
+ * Represents a type of authorization grant.
+ *
+ * An authorization grant is a credential representing the resource
+ * owner's authorization (to access its protected resources) used by the
+ * client to obtain an access token.  OAuth 2.0 defines four
+ * grant types -- authorization code, implicit, resource owner password
+ * credentials, and client credentials -- as well as an extensibility
+ * mechanism for defining additional types.
+ *
+ * @link http://tools.ietf.org/html/rfc6749#section-1.3 Authorization Grant (RFC 6749, §1.3)
```
shadowhand: Awesome!

src/Token/AccessToken.php
ramsey: I tend to agree, but I go back-and-forth on my own arguments on this.

As it is, it took me a while to write up these descriptions, since I had to sift through the code to make sense of it. In fact, looking at the code, you can see that you're able to pass in `grantFactory`, `requestFactory`, `httpClient`, and `randomFactory` as options, too, and they are properly set, but then they are overridden if they aren't provided as collaborators, so I listed them in the comment to make it clear to implementors where they should go.

PR #401, opened 2015-08-12
Pass ResponseInterface to IdentityProviderException

ramsey:
I came across this while updating class docblock information.

Previously, `checkResponse()` passed `$data` to `IdentityProviderException`. However, this looks incorrect, since the third parameter for `IdentityProviderException` is labeled as `$response` and its `getResponseBody()` method says it returns a `ResponseInterface`. So, it seemed to me that passing in `$data` was incorrect, but I'm not sure, so I'll need some eyes on this to tell me if this is the intended behavior.

If accepted, this will also type-hint the `$response` parameter on `IdentityProviderException` as a `ResponseInterface` to make it clearer what it expects.

Should the `getResponseBody()` method name be changed to `getResponse()`, since it's not returning a body but a full response?

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -679,7 +679,7 @@ abstract class AbstractProvider
      *
      * @throws IdentityProviderException
      * @param  ResponseInterface $response
-     * @param  string $data Parsed response data
+     * @param  array|string $data Parsed response data
      * @return void
      */
     abstract protected function checkResponse(ResponseInterface $response, $data);
--- a/src/Provider/Exception/IdentityProviderException.php
+++ b/src/Provider/Exception/IdentityProviderException.php
@@ -14,6 +14,8 @@
 
 namespace League\OAuth2\Client\Provider\Exception;
 
+use Psr\Http\Message\ResponseInterface;
+
 /**
  *
  */
@@ -24,7 +26,12 @@ class IdentityProviderException extends \Exception
      */
     protected $response;
 
-    public function __construct($message, $code, $response)
+    /**
+     * @param string $message
+     * @param int $code
+     * @param ResponseInterface $response
+     */
+    public function __construct($message, $code, ResponseInterface $response)
     {
         $this->response = $response;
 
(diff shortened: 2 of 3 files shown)


ramsey:
Thoughts @shadowhand, @rtheunissen?

shadowhand:
The IDPE should get the response body as an array, not the entire response object. That's how it was intended to work, so far as I can tell. And I think this would constitute an API change and can't be made until v1.1.

ramsey:
Thanks for the clarification.

I'll update the pull request to reflect this, so that I can make sure I'm clear on the intent.

ramsey:
My latest changes don't change the interface. They update the documentation to make it clear that `$data` could be an array or a string. You'll recall that `AbstractProvider::parseResponse()` may return a string if it can't parse the response as JSON or urlencoded values. So, it's possible that `IdentityProviderException` could receive data as a string of the response body, rather than an array.

shadowhand:
:+1:

rtheunissen:
Time zones make it difficult for me to contribute to these conversations before they are concluded :(

@shadowhand is right though, the data array is correct and I missed that in the docblocks.

stevenmaguire:
:+1: for scrutinizer!

ramsey:
Sorry about the timezones, @rtheunissen. :-)

I'm all for using Scrutinizer. I've tried to set it up in the past, but I got bogged down in all the configuration, so I set it aside and never came back to it. Anyone want to help set it up?

rtheunissen:
Sure thing! I'm very familiar with it so I'll set it up for us. PSR-2 and doc checks.

GawainLynch:
Lurking ninja chiming in here… Scrutinizer can give you a hard time on some occasions, but more often than not shows up complete derps that everyone signed off on… Please do @rtheunissen :+1:

rtheunissen:
@GawainLynch absolutely right. Have to filter through the overly pedantic or incorrect analyses.

stevenmaguire:
I think its a great tool. My chief complain is the inconsistent communication behavior when performing analysis and waiting for input from third-party services, like Travis. I've had several occasions of "no word yet, so we gave up" and "we heard that the tests failed, so your build is broken" even though there's only been on build and it succeeded. Those bits are annoying sometimes, and its still worth it.

rtheunissen:
Someone with admin access will need to set it up. I'll create an issue for it.

PR #403, opened 2015-08-12
Clean up documentation with working examples and better description

ramsey:


composer.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "league/oauth2-client",
-    "description": "OAuth 2.0 Client Library",
+    "description": "This package makes it simple to integrate your application with OAuth 2.0 service providers.",
```
shadowhand: The rest looks really good, but this seems a bit wordy. Perhaps:

> A package to simplify integrating with OAuth 2.0 service providers

I don't think a period on the end is necessary, either.

ramsey: Fair enough.

PR #404, opened 2015-08-12
[WIP] Main site initial implementation

tpavlek:
This is an initial implementation of the gh-pages branch to better document the client. For the most part, it was taking the information that already exists in the READMEs, adjusting some of it to fit the site format, and putting it in.

I'd probably suggest once the site is completed, we decommission some of the READMEs to avoid having to maintain information in two places. The one caveat with that is it makes it more difficult for less-experienced contributors to submit a pull with their new provider implementation - maybe we'll just have to have them file an issue instead?

Another thing I'm really unhappy with with the documentation right now is the usage example. It's fine, I suppose, but I feel the general user of this package is probably using a framework, and the usage example is framed as a "single script", which can be hard to contextualize for new developers using something like Laravel, which has a very expressive and developer-friendly syntax. Are we opposed to adding a page that adds a more contextual example eg. "Create a redirect method in your controller, that returns `$provider->getAuthorizationUrl()`"? I think there's probably a lot of value to that.

### DIFF
--- a/_data/menu.yml
+++ b/_data/menu.yml
@@ -1,2 +1,8 @@
 Getting Started:
     Introduction: '/'
+    "Basic Usage": '/usage/'
+
+Providers:
+    "Official Providers": '/providers/league/'
+    "Third-Party Providers": '/providers/thirdparty/'
+    "Implementing Your Own": '/providers/implementing/'
--- a/_data/project.yml
+++ b/_data/project.yml
@@ -1,5 +1,5 @@
-title:
-tagline:
+title: OAuth 2 Client
+tagline: Easy integration with OAuth2 service providers
 description:
 google_analytics_tracking_id:
 base_href:
--- a/index.md
+++ b/index.md
@@ -1,8 +1,40 @@
 ---
 layout: default
 permalink: /
-title: Introduction
+title: oauth2-client
 ---
 
-Introduction
-============
+League/oauth2-client
+======================
+
+[![Gitter Chat](https://img.shields.io/badge/gitter-join_chat-brightgreen.svg?style=flat-square)](https://gitter.im/thephpleague/oauth2-client)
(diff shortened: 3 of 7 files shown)


ramsey:
I think @shadowhand is working on some more detailed demos that show usage within a framework context. In general, most end-developers should not be using this library but, rather, one of its descendant packages.

My fear with showing examples in a framework context is that you then have all the nuances of whatever framework is used that become stumbling blocks to grasping the concepts of the OAuth client library. From a teaching standpoint, this is bad, and it will also lead to developers seeing this as a library for use with X framework and not Y framework, when the reality is that it can be used in both.

As a library user, I like having documentation in the package and not on a separate website. Maybe I'm in the minority here, but I do agree that we shouldn't duplicate documentation on both the website and in the READMEs. We should put it in one place and link to it from the other. I believe @reinink has done some interesting things using GitHub's version of Jekyll in order to display data on a gh-pages website using information from the repo, so maybe he has some pointers.

FYI, I updated the README as you were submitting this PR. Sorry!

tpavlek:
I thought the README update would be coming given that there was an issue asking about client_credentials, and that was a generally undocumented flow :)

I think you're probably right about framework-specific stuff. Perhaps just a separate section detailing in prose what is happening and what lines do what will allow an end-user to more accurately determine how each method call fits in with the context of his application. My open source cycles are done for the next couple days, but after that I might draft something up to see how it looks if someone else hasn't.

ramsey:
Sounds good. Should I go ahead and merge this in or leave it as a WIP?

The more I think about it, the more I like the idea of pulling the examples and provider list out of the master branch and putting it into the gh-pages branch and linking to it from the README.

tpavlek:
I think WIP until its feature-complete in terms of examples.

shadowhand:
@ramsey I'd appreciate if you merged this as-is. Something is better than nothing and it would allow others (see my mistake in #412) to start working from a this as a baseline.

tpavlek:
+1

My initial recommendation was ill thought through

PR #407, opened 2015-08-14
Small fixed suggested by Scrutinizer CI

rtheunissen:
This PR aims to fix a few small issues that Scrutinizer pointed out.

Does not break BC.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -420,7 +420,7 @@ abstract class AbstractProvider
     /**
      * Returns the key used in the access token response to identify the resource owner.
      *
-     * @return string Resource owner identifier key
+     * @return string|null Resource owner identifier key
      */
     protected function getAccessTokenResourceOwnerId()
     {
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -82,7 +82,7 @@ abstract class AbstractProvider
     protected $requestFactory;
 
     /**
-     * @var HttpAdapterInterface
+     * @var HttpClientInterface
      */
     protected $httpClient;
 
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -707,7 +707,7 @@ abstract class AbstractProvider
      */
     protected function prepareAccessTokenResponse(array $result)
     {
-        if ($this->getAccessTokenResourceOwnerId()) {
+        if ($this->getAccessTokenResourceOwnerId() !== null) {
             $result['resource_owner_id'] = $this->getValueByKey(
                 $this->getAccessTokenResourceOwnerId(),
                 $result
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -442,7 +442,7 @@ abstract class AbstractProvider
      * Checks that a provided grant is valid, or attempts to produce one if the
      * provided grant is a string.
      *
-     * @param  mixed $grant
+     * @param  AbstractGrant|string $grant
      * @return AbstractGrant
      */
     protected function verifyGrant($grant)
--- a/src/Provider/GenericProvider.php
+++ b/src/Provider/GenericProvider.php
@@ -141,7 +141,8 @@ class GenericProvider extends AbstractProvider
     private function assertRequiredOptions(array $options)
     {
         $missing = array_diff_key(array_flip($this->getRequiredOptions()), $options);
-        if ($missing) {
+
+        if (!empty($missing)) {
             throw new InvalidArgumentException(
                 'Required options not defined: ' . implode(', ', array_keys($missing))
             );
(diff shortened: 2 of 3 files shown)


src/Provider/AbstractProvider.php
```diff
@@ -82,7 +82,7 @@
     protected $requestFactory;
 
     /**
-     * @var HttpAdapterInterface
+     * @var HttpClientInterface
```
shadowhand: Good catch.

src/Provider/AbstractProvider.php
```diff
@@ -707,7 +707,7 @@ protected function parseResponse(ResponseInterface $response)
      */
     protected function prepareAccessTokenResponse(array $result)
     {
-        if ($this->getAccessTokenResourceOwnerId()) {
+        if ($this->getAccessTokenResourceOwnerId() !== null) {
```
shadowhand: This doesn't seem right, because `false` and `''` would not be valid possibilities.

src/Provider/GenericProvider.php
```diff
@@ -141,7 +141,8 @@ protected function getRequiredOptions()
     private function assertRequiredOptions(array $options)
     {
         $missing = array_diff_key(array_flip($this->getRequiredOptions()), $options);
-        if ($missing) {
+
+        if (!empty($missing)) {
```
shadowhand: This gets us nothing. The two statements are effectively equivalent.

src/Tool/MacAuthorizationTrait.php
```diff
@@ -26,7 +26,7 @@
     /**
      * Returns the id of this token for MAC generation.
      *
-     * @param  AccessToken $token
+     * @param  AccessToken|null $token
```
shadowhand: Type hint does not allow the token to be `null` and the method clearly needs a valid token.

src/Tool/MacAuthorizationTrait.php
```diff
@@ -53,8 +53,7 @@
     /**
      * Returns the authorization headers for the 'mac' grant.
      *
-     * @param  AccessToken $token
-     *
+     * @param  AccessToken|null $token
```
shadowhand: Same here.

src/Provider/AbstractProvider.php
ramsey: The specific Scrutinizer message is:

> The expression `$this->getAccessTokenResourceOwnerId()` of type `string|null` is loosely compared to true; this is ambiguous if the string can be empty. You might want to explicitly use `!== null` instead.

I'm not sure how Scrutinizer thinks the function returns `string|null`, when the return type annotation for it clearly shows just `string`. We probably just want to ignore this Scrutinizer issue, or figure out a way to annotate the return type hint to show that it can be (and might be) an empty string.

Clearly, if it's an empty string, we don't want this check to pass, so `!== null` won't work for us.

src/Provider/GenericProvider.php
ramsey: Again, Scrutinizer is being over-zealous here:

> The expression `$missing` of type `array` is implicitly converted to a boolean; are you sure this is intended? If so, consider using `! empty($expr)` instead to make it clear that you intend to check for an array without elements.

It should be fine to leave as is, and we avoid a function call, if you want an optimization argument. ;-)

src/Tool/MacAuthorizationTrait.php
ramsey: This stems from this Scrutinizer message:

> It seems like `$token` defined by parameter `$token` on line 65 can be `null`; however, `League\OAuth2\Client\Tool\MacAuthorizationTrait::getTokenId()` does not accept `null`, maybe add an additional type check?

The signature for the method in which this suggestion occurs is:

    protected function getAuthorizationHeaders($token = null)

This method calls `getTokenId()`, which doesn't allow a `null` token to be passed in (according to its return type annotation). Perhaps the solution is to not default to `null` for `getAuthorizationHeaders()`.

shadowhand: Dropping the `= null` default makes sense to me. [This condition already prevents it](https://github.com/thephpleague/oauth2-client/blob/1b0fe447ae474bcfd83f3f8578ce45f5077f2630/src/Provider/AbstractProvider.php#L811-L816) so we lose nothing.

src/Provider/AbstractProvider.php
rtheunissen: > I'm not sure how Scrutinizer thinks the function returns string|null

Maybe because the default value is `null`? ie. 

`const ACCESS_TOKEN_RESOURCE_OWNER_ID = null;`

The rationalisation behind this change is that if the method doesn't return null, that means the provider explicitly specified *some* value.

src/Provider/GenericProvider.php
rtheunissen: > The two statements are effectively equivalent.

Yup, effectively equivalent, but not expressively.

> It should be fine to leave as is, and we avoid a function call, if you want an optimization argument.

Sure thing. Weak, but valid. ;)

src/Tool/MacAuthorizationTrait.php
rtheunissen: Sounds good to me, I'll amend.

ramsey:
Merging in for 1.0.0. We can continuing discussing the changes recommended by Scrutinizer, but overall, these don't change functionality.

PR #408, opened 2015-08-14
[2.0] Make getResponse consistent with getRequest

rtheunissen:
I realise that this breaks BC, so it would have to be a part of the next major version.

This came up when I was looking at [an issue that Scrutinizer pointed out](https://scrutinizer-ci.com/g/thephpleague/oauth2-client/issues/master/files/src/Provider/AbstractProvider.php?orderField=path&order=asc#inspectioncomment-1186598870), where the current `getResponse` returns a _parsed_ response, which could be a string or an array (or null?). But the required type for the response parameter in `prepareAccessTokenResponse` is only `array`.

It doesn't _realistically_ make sense for an access token response to **not** be an array, but from a purely correct and well documented code point of view, this seems inconsistent. It's not clear that getResponse returns a parsed response, where getRequest returns a `RequestInterface`.

If for whatever reason the value passed to `prepareAccessTokenResponse` is not an array, I think an `UnexpectedValueException` is more meaningful than the default PHP error.

**Proposed solution**:
- Rename `getResponse` to `getParsedResponse`
- Rename `sendRequest` to `getResponse`

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -533,7 +533,7 @@ abstract class AbstractProvider
 
         $params   = $grant->prepareRequestParameters($params, $options);
         $request  = $this->getAccessTokenRequest($params);
-        $response = $this->getResponse($request);
+        $response = $this->getParsedResponse($request);
         $prepared = $this->prepareAccessTokenResponse($response);
         $token    = $this->createAccessToken($prepared, $grant);
 
@@ -596,7 +596,7 @@ abstract class AbstractProvider
      * @param  RequestInterface $request
      * @return ResponseInterface
      */
-    protected function sendRequest(RequestInterface $request)
+    protected function getResponse(RequestInterface $request)
     {
         try {
             $response = $this->getHttpClient()->send($request);
@@ -610,11 +610,11 @@ abstract class AbstractProvider
      * Sends a request and returns the parsed response.
      *
      * @param  RequestInterface $request
-     * @return mixed
+     * @return array|string
      */
-    public function getResponse(RequestInterface $request)
+    public function getParsedResponse(RequestInterface $request)
     {
-        $response = $this->sendRequest($request);
+        $response = $this->getResponse($request);
         $parsed = $this->parseResponse($response);
 
         $this->checkResponse($response, $parsed);
@@ -660,7 +660,7 @@ abstract class AbstractProvider
      *
      * @throws UnexpectedValueException
      * @param  ResponseInterface $response
-     * @return array
+     * @return array|string
      */
     protected function parseResponse(ResponseInterface $response)
     {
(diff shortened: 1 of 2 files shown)


src/Provider/AbstractProvider.php
```diff
@@ -702,18 +702,26 @@ protected function parseResponse(ResponseInterface $response)
...
      * parent method when overloading this method.
      *
-     * @param  mixed $result
-     * @return array
+     * @param  array|string $response
+     * @return array|string
+     * @throws UnexpectedValueException
      */
-    protected function prepareAccessTokenResponse(array $result)
+    protected function prepareAccessTokenResponse($response)
```
shadowhand: This change doesn't make sense to me. The OAuth2 spec clearly states what the responses will be, and it is always a hash of keys and values.

ramsey: It's possible for `parseResponse()` to return a string without throwing an exception (if it couldn't parse the response as JSON or urlencoded form values). After doing so, the `getResponse()` method (as it is currently named, not as its name changes to in this PR) passes the string to `checkResponse()`. The `checkResponse()` method is implemented by the provider, so the provider could choose to throw an exception if it's not a proper array, or not if they want the string, instead.

As it stands now, `checkResponse()` can return a string value, which gets passed back up the chain and then into the next call to `prepareAccessTokenResponse()`, where it will throw an error for not being an array.

Should `parseResponse()` or `checkResponse()` always ensure the value is an array before passing it back up the chain?

shadowhand: I don't think so. The key thing to remember here is that request/response is more than just the OAuth flow, since we also have resource owners. In the case of `prepareAccessTokenResponse`, we absolutely know that the response must be an array. In other situations, requesting other API endpoints, that may not always be the case and we have a mixed result.

In this case, Scrutinizer simply doesn't have enough facts to make a judgement call about the code path in question. The warning should be ignored, because we can reason about the code path more completely than an automated test.

ramsey:
Yes, this breaks BC and `getResponse()` is a public method, so I'm going to think over this for a bit more before deciding whether to reject or accept it.

In general, I like the changes, but I need to look at some of the providers to see how many of them use these methods. If no one uses them (or only a couple do), then maybe it's safe to make this change now.

shadowhand:
I'm of the opinion that this PR is churn and doesn't represent any changes of real value. All it does is modify method names, since the type hint change is not necessary, as per discussion.

src/Provider/AbstractProvider.php
rtheunissen: > In the case of prepareAccessTokenResponse, we absolutely know that the response must be an array

I agree with this, as well as everything else both you and @ramsey are saying. We know that it has to be an array, and the typehint should enforce that.

> Type hinting provides the same protections that throwing an exception does.

"If for whatever reason the value passed to prepareAccessTokenResponse is not an array, I think an UnexpectedValueException is more meaningful than the default PHP error."

rtheunissen: > Should parseResponse() or checkResponse() always ensure the value is an array before passing it back up the chain?

I don't think so, as they could both be used outside of the access token request (I think).

rtheunissen:
> I'm of the opinion that this PR is churn and doesn't represent any changes of real value. All it does is modify method names, since the type hint change is not necessary, as per discussion.

Why should a change be significant for it to be valuable? Everyone has a different definition for 'real value'. Renaming a method to be more inline with what it does and the behaviour one can expect from it is 'real value' to me. I am aware that it breaks BC, which is why we have semver.

rtheunissen:
> I'm going to think over this for a bit more before deciding whether to reject or accept it

:+1:

shadowhand:
> I am aware that it breaks BC, which is why we have semver.

I feel like you are suggesting that we skip 1.0 and go directly to 1.1 to "fix" a couple of method names that are only questionably incorrect. To me, this is a perfect example of churn.

rtheunissen:
> I feel like you are suggesting that we skip 1.0 and go directly to 1.1

Include it when 1.1 is eventually released anyway, as it's not an important fix.

ramsey:
This has been open for over a year. Is there any interest in working this into a 2.0 version of the library? If so, perhaps we introduce the new method names as proxies for the current methods and deprecate the use of the current methods in a subsequent 1.x release and then message all the third-party providers by posting "upgrade" issues on their repos.

shadowhand:
I can take this on. Having a method that returns the unparsed response would be useful when using the client as an API gateway (which it is actually very good at).

shadowhand:
This will be addressed with #587. Would appreciate your review. 😅

PR #417, opened 2015-08-27
Added Slack to providers

adam-paterson:


shadowhand:
Some of the type hinting is missing but generally seems to pretty good. I would recommend using the [auth.test](https://api.slack.com/methods/auth.test) endpoint to get user information, but it will require overloading more of AbstractProvider to make the additional call. :)

adam-paterson:
@shadowhand, I did use `auth.test` originally but thought you'd recommend using a resource that actually returns some user information. I'll swap it out anyway and add the missing typehints.

shadowhand:
@adam-paterson you would need to use `auth.test` and then `users.info` to fetch the resource owner details.

adam-paterson:
@shadowhand still working on this, had other commitments at the moment.

adam-paterson:
@shadowhand I've updated this now, can you check I've taken the right approach?

shadowhand:
@adam-paterson looking better! I did notice one issue with the query strings, which I reported at adam-paterson/oauth2-slack#2 but otherwise the new implementation for fetching resource owner is great.

adam-paterson:
@shadowhand thanks for that, fixed now.

PR #421, opened 2015-09-04
Rdio provider

adam-paterson:


tpavlek:
Make sure to update the providers list in the gh-pages branch, until we can
figure out a solution for mirroring them

shadowhand:
The `RdioResourceOwner` would probably be improved by not extending `GenericResourceOwner`, but otherwise looks quite good. :+1:

adam-paterson:
@tpavlek @shadowhand I'll make those two changes

PR #430, opened 2015-09-08
Minor fixes

adam-paterson:
* Permissions on AccessToken.php changed to a-x-  
* ~~User specific files in .gitignore~~
* Removed unused use statements


rtheunissen:
:+1: for removing unused use statements. 

I don't see how those .gitignore statements were a problem. What's your rationale for removing them?

adam-paterson:
@rtheunissen I've always been under the impression it's best practice to keep the .gitignore clean as possible. For user specific files which will occur on every project (ide configs, etc) they should be declared in a [global .gitignore] (https://help.github.com/articles/ignoring-files/#create-a-global-gitignore)

rtheunissen:
Sounds good to me, I should have done that a long time ago.

ramsey:
I don't understand the value of removing the environment-specific lines from `.gitignore`. In the past, it's been explained to me that they're even "dangerous" to leave there, but no one has given me good enough reason to remove them.

As someone who develops on multiple environments with multiple VMs, it's annoying to have to maintain my global `.gitignore` file across environments, so I think it's logical to keep these in the project, and since this is usually a relatively small file anyway, I don't see how it makes  difference.

I'm open to hearing arguments for removing these lines, though. I'd like to be swayed. :-)

tpavlek:
I tend to agree with @ramsey.

Wherever I hear this it feels like bikeshedding. Keeping one line in the
.gitignore is worth it if it allows a new contributor to help out and we
don't have to send the pull back because he "didn't set up his environment
right".

I'm for removing friction even at the expense of purity

shadowhand:
I think it is silly to have platform-specific things in a project `.gitignore`. A global `.gitignore` is much better for those kinds of things. That said, having them in a project hurts no one.

rtheunissen:
I agree with @ramsey as well even though I'm usually in favour of the 'correct' or 'proper' way to do things. 

"Pragmatism, not idealism" - Radiohead

adam-paterson:
I still believe that in a perfect world the .gitignore should only include project specific files that aren't required to be tracked. However, contributions to OS outweigh everybody having the same environment.

Amended the commit.

PR #434, opened 2015-09-22
Allow access tokens to be created from storage

### LINKED ISSUE #431: Restore AccessToken
I am store the data from AccessToken in $_SESSION.

When I want restoring AccessToken from $_SESSION
```
$accessToken = new AccessToken($_SESSION['accessToken']);
```
I can have wrong expires-field, and method hasExpires() returns «true» always.

For example, if user use access_token after expired time, constructor writes to expires-filed 
```
time()+«time_from_SESSION»
```
I mean it's wrong

shadowhand:
Previously if an access token expiration was in the past, it was assumed that the `expires` value was a second value and would make the token appear to not be expired.

To prevent this discrepancy, allow for any past timestamp after October 2012 to be used.

Fixes #431

### DIFF
--- a/test/src/Token/AccessTokenTest.php
+++ b/test/src/Token/AccessTokenTest.php
@@ -32,6 +32,19 @@ class AccessTokenTest extends \PHPUnit_Framework_TestCase
         $this->assertLessThan(time() + 200, $expires);
     }
 
+    public function testExpiresPastTimestamp()
+    {
+        $options = ['access_token' => 'access_token', 'expires' => strtotime('5 days ago')];
+        $token = $this->getAccessToken($options);
+
+        $this->assertTrue($token->hasExpired());
+
+        $options = ['access_token' => 'access_token', 'expires' => 3600];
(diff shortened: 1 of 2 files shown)


test/src/Token/AccessTokenTest.php
```diff
@@ -32,6 +32,19 @@ public function testExpiresInCorrection()
         $this->assertLessThan(time() + 200, $expires);
     }
 
+    public function testExpiresPastTimestamp()
+    {
+        $options = ['access_token' => 'access_token', 'expires' => strtotime('5 days ago')];
+        $token = $this->getAccessToken($options);
+
+        $this->assertTrue($token->hasExpired());
```
shadowhand: This test fails without the patch to `AccessToken`. :)

ramsey:
I just merged in something before this, which created conflicts. Would you mind rebasing please?

shadowhand:
@ramsey done. Should be good now.

ramsey:
Thanks!

PR #435, opened 2015-09-22
Show code coverage results in CLI

shadowhand:
Having the coverage stored in a file is not very useful.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -15,8 +15,8 @@
 	<logging>
         <log type="tap" target="build/report.tap"/>
         <log type="junit" target="build/report.junit.xml"/>
+        <log type="coverage-text" target="php://stdout"/>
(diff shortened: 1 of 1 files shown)


phpunit.xml
```diff
@@ -15,8 +15,8 @@
 	<logging>
         <log type="tap" target="build/report.tap"/>
         <log type="junit" target="build/report.junit.xml"/>
+        <log type="coverage-text" target="php://stdout"/>
         <log type="coverage-html" target="build/coverage" charset="UTF-8" yui="true" highlight="true"/>
-        <log type="coverage-text" target="build/coverage.txt"/>
```
shadowhand: @ramsey any reason it was being stored in a file?

ramsey: I never noticed that. Not something I did, and I doubt there's a good reason to keep it in a file. :-)

PR #442, opened 2015-09-25
Add Foursquare provider to readme

stevenmaguire:


shadowhand:
@stevenmaguire you're quite the collector!

stevenmaguire:
@shadowhand just doing my part! It's pretty fast to put them together. It's fun learning about the ins and outs of each service too.

PR #447, opened 2015-10-03
Yahoo provider is added to thirdparty.md

hayageek:
Hi,
I have added yahoo provider for oauth2-client https://github.com/hayageek/oauth2-yahoo


ramsey:
Thanks!

PR #450, opened 2015-10-05
Add ArrayAccessorTrait

stevenmaguire:
Previously the AbstractProvider included a protected method to parse a given array for values using the dot notation. This code is pretty generic and in building third party providers, I've discovered that it needs to be repeated for certain ResourceOwner classes. I am proposing to move this dot notation method to a new ArrayAccessorTrait so that it can serve the current internal purpose as well as serve third party purposes.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -22,6 +22,7 @@ use League\OAuth2\Client\Grant\GrantFactory;
 use League\OAuth2\Client\Provider\Exception\IdentityProviderException;
 use League\OAuth2\Client\Token\AccessToken;
 use League\OAuth2\Client\Tool\RequestFactory;
+use League\OAuth2\Client\Tool\ArrayAccessorTrait;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 use RandomLib\Factory as RandomFactory;
@@ -34,6 +35,8 @@ use UnexpectedValueException;
  */
 abstract class AbstractProvider
 {
+    use ArrayAccessorTrait;
+
     /**
      * @var string Key used in a token response to identify the resource owner.
      */
@@ -707,8 +710,8 @@ abstract class AbstractProvider
     {
         if ($this->getAccessTokenResourceOwnerId() !== null) {
             $result['resource_owner_id'] = $this->getValueByKey(
-                $this->getAccessTokenResourceOwnerId(),
-                $result
+                $result,
+                $this->getAccessTokenResourceOwnerId()
             );
         }
         return $result;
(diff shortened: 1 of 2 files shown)


shadowhand:
I wonder if it would be worthwhile to just use [jmespath](https://github.com/jmespath/jmespath.php) instead?

stevenmaguire:
Yeah, I suppose that would work too. One downside is an additional dependancy.

I could go either way. I wish this low-level functionality was native to PHP right now!

shadowhand:
I could really go either way. I just feel that having any array tool is going to lead to more and more complexity being added to it and JMESpath is a standard way to do just about anything without reinventing the wheel.

stevenmaguire:
What are your thoughts on keeping the Array tool class and using the JMESpath library as an implementation of "package methods"? Adding official functionality to this package's API still.

shadowhand:
I don't see that as being terribly useful. It just wraps the class for no particular reason, imo.

stevenmaguire:
The benefit I see in a solution where this package exposes a single public static method, in this new class or on the Abstract Provider, is that the third party providers can access this single piece of functionality without a new dependency. I do agree that there is a concern of reinventing the wheel and the reality is that this package only needs this one method currently, and it already exists. I would like to expose it to the rest of the package as a utility and allow third party providers to utilize it. 

I think one of these three options would accomplish this:

1) Merge the current proposed solution.
2) Update the AbstractProvider::getValueByKey method to be public static.
3) Update the current proposed solution to use JMESpath as the implementation, such that this package maintains the dependency.

src/Tool/ArrayNavigator.php
```diff
@@ -0,0 +1,52 @@
...
+ * @link https://packagist.org/packages/league/oauth2-client Packagist
+ * @link https://github.com/thephpleague/oauth2-client GitHub
+ */
+
+namespace League\OAuth2\Client\Tool;
+
+/**
+ * Provides generic array navigation tools.
+ */
+class ArrayNavigator
```
shadowhand: It might make more sense to have this be `trait ArrayAccessorTrait`

src/Tool/ArrayNavigator.php
```diff
@@ -0,0 +1,52 @@
...
+{
+    /**
+     * Returns a value by key using dot notation.
+     *
+     * @param  array      $data
+     * @param  string     $key
+     * @param  mixed|null $default
+     * @return mixed
+     */
+    public static function getValueByKey(array $data, $key, $default = null)
```
shadowhand: with `private function getValueByKey`

src/Tool/ArrayNavigator.php
stevenmaguire: I considered that as well. I think that would work too! I waffled over the trait v. static class for a bit!

stevenmaguire:
I've updated this solution to use a trait instead. I think it will accomplish my goals and satisfy some of the, great and valid, concerns @shadowhand raised!

shadowhand:
:+1:

ramsey:
Looks good. Thanks!

PR #456, opened 2015-11-13
Add Mollie provider

Daanvm:
I created a provider for [Mollie](https://www.mollie.com/en).


shadowhand:
I'd like to see your provider follow [PSR-2 style](http://www.php-fig.org/psr/psr-2/) but I don't think we require it.

:+1:

PR #461, opened 2015-11-16
Add Azure Active Directory provider to thirdparty providers

hajekj:
Azure Active Directory is very important identity provider in the business segment. This library for oauth2-client enables the use of AAD.


shadowhand:
It would be nice if this provider would follow [PSR-2](http://www.php-fig.org/psr/psr-2/) but we do not require it.

shadowhand:
I've opened a number of issues to address basic functionality.

hajekj:
Thank you very much, working on those right now.

hajekj:
I believe I have fixed those issues. I am going to work on PSR-2 compliance in future.

PR #470, opened 2015-12-01
Add Shotbow as a provider

navarr:


stevenmaguire:
Thanks for contributing @navarr! Typically the custom providers follow the naming convention of `{vendor}/oauth2-{provider}`. Your package appears to be called `shotbow/oauth2-client`, where `oauth2-client` is the same name as the base package. If it's all the same to you, would you mind renaming the package to `shotbow/oauth2-shotbow`?

navarr:
Change has been made as requested.

stevenmaguire:
:+1:

PR #471, opened 2015-12-10
Add tests for array accessor trait, fixed one bug

stevenmaguire:


src/Tool/ArrayAccessorTrait.php
```diff
@@ -27,7 +27,7 @@
      * @param  mixed|null $default
      * @return mixed
      */
-    private function getValueByKey(array $data, $key, $default = null)
+    protected function getValueByKey(array $data, $key, $default = null)
```
shadowhand: Why protected?

src/Tool/ArrayAccessorTrait.php
```diff
@@ -37,7 +37,7 @@ private function getValueByKey(array $data, $key, $default = null)
             $keys = explode('.', $key);
 
             foreach ($keys as $innerKey) {
-                if (!array_key_exists($innerKey, $data)) {
+                if (!is_array($data) || !array_key_exists($innerKey, $data)) {
```
shadowhand: This is impossible because the type hint requires `$data` to be an array.

stevenmaguire: `$data` is being changed during this loop on line 44

src/Tool/ArrayAccessorTrait.php
stevenmaguire: I don't know what you are talking about ;)

src/Tool/ArrayAccessorTrait.php
shadowhand: Ah, good call.

stevenmaguire: The basic test at https://github.com/stevenmaguire/oauth2-client/blob/add-array-accessor-tests/test/src/Tool/ArrayAccessorTraitTest.php#L40 broke it because the dot notation expects two layers and `$data` was set to the string value of the first layer. I don't know how I missed this previously and I am glad I caught it!

PR #473, opened 2015-12-30
Include resource owner id in access token json

### LINKED ISSUE #472: AccessToken::jsonSerialize does not return resourceOwnerId
`jsonSerialize()` method on `League\OAuth2\Client\Token\AccessToken` does not return `resourceOwnerId` and thus not consistent.

shadowhand:
Fixes #472


ramsey:
:+1:

PR #474, opened 2015-12-30
added checkAccessTokenResponse abstract function

aporat:
some services (atleast Github & Instagram as far as I noticed) have different error response format when a request is made to exchange the auth code for an access token.

this pull request adds a new abstract function named `checkAccessTokenResponse` which is responsible to handling the access token response and making sure it's a valid response.

this is a breaking change, as any provider will have to implement this abstract function.

Is there another solution for this issue? both providers I tried using (github & instagram) don't handle the case when an invalid auth code is used. the providers will just throw a `new InvalidArgumentException('Required option not passed: "access_token"')` exception, as the `checkResponse` doesn't recognize the error response

for example, see https://developer.github.com/v3/oauth/

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -536,7 +536,7 @@ abstract class AbstractProvider
 
         $params   = $grant->prepareRequestParameters($params, $options);
         $request  = $this->getAccessTokenRequest($params);
-        $response = $this->getResponse($request);
+        $response = $this->getAccessTokenResponse($request);
         $prepared = $this->prepareAccessTokenResponse($response);
         $token    = $this->createAccessToken($prepared, $grant);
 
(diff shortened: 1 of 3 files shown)


shadowhand:
cc @stevenmaguire as the maintainer of both providers.

stevenmaguire:
Thanks for the heads up @shadowhand. 

@aporat each of the providers implements their own logic for "checking responses" and I can say that both the Instagram and Github providers have a basic check in place based on each of the APIs error handling documentation. I would encourage you to take a look at the `checkResponse` methods within each of the providers and propose an update to that method that addresses your concern here. 

I notice that you have already created a pull request for the Instagram provider.  https://github.com/thephpleague/oauth2-instagram/pull/8

I recommend closing this PR without merging. These should be addressed in the individual providers unless a more consistent concern is raised.

aporat:
thanks. closed.

PR #477, opened 2016-01-05
Fix resource_owner_id being put into the wrong array

Lavoaster:
Ya'll should probably check scrutinzer some more. It highlighted this issue, albeit not the severity, when it came up. 

https://scrutinizer-ci.com/g/thephpleague/oauth2-client/inspections/044f7d1f-3923-4b78-8c04-313390c0594a/issues/files/src/Token/AccessToken.php?status=new&orderField=path&order=asc&honorSelectedPaths=0

### DIFF
--- a/src/Token/AccessToken.php
+++ b/src/Token/AccessToken.php
@@ -190,7 +190,7 @@ class AccessToken implements JsonSerializable
         }
 
         if ($this->resourceOwnerId) {
-            $params['resource_owner_id'] = $this->resourceOwnerId;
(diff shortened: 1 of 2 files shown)


shadowhand:
:+1: please add a unit test so this gets caught!

Lavoaster:
Re-did the commit with tests, also renamed the associated test itself since it was misspelt. (I can totally talk, putting a typo in the original commit!)

ramsey:
Thanks!

PR #483, opened 2016-02-06
Add Eve Online provider

okwinza:


shadowhand:
I checked out the provider in question, it looks to be of high quality. :+1:

okwinza:
Thanks @shadowhand ! 
I used official LinkedIn provider as a reference so it's mostly copy&paste :D 
Although code coverage looks strange to me because codebase is pretty much the same yet there is huge difference in % covered. 

Would appreciate any advice on this.

stevenmaguire:
@okwinza that LinkedIn package _is_ pretty good, if I say so myself ;)

Your code coverage is so poor because your package includes an `example` directory, with what I assume to be code samples. That example directory is not excluded from the test suite and there are no tests for the code in that directory. If you exclude this directory from your test suite you will be in good shape on the code coverage front.

okwinza:
@stevenmaguire thanks, got all 100 after excluding that dir.

stevenmaguire:
:+1:

PR #484, opened 2016-02-09
Add Fitbit provider to readme

djchen:


shadowhand:
The provider itself is of high quality, but the current `FitbitUser` implementation is very limited and does not return all of the user information in the `toArray` method. I have opened an issue about this.

PR #485, opened 2016-02-10
Enable dynamic parameters being passed into the authorization URL

hajekj:
When for example using Azure AD, you may need to pass additional parameters to the authorization URL - like `prompt`. This change enables passing custom parameters to the authorization URL as per each provider's needs. More options for Azure AD can be found [here](https://msdn.microsoft.com/en-us/library/azure/dn645542.aspx)
For Google those are for example `access_type` or `login_hint` - found [here](https://developers.google.com/youtube/v3/guides/auth/server-side-web-apps)

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -332,14 +332,11 @@ abstract class AbstractProvider
         // Store the state as it may need to be accessed later on.
         $this->state = $options['state'];
 
-        return [
-            'client_id'       => $this->clientId,
-            'redirect_uri'    => $this->redirectUri,
-            'state'           => $this->state,
-            'scope'           => $options['scope'],
-            'response_type'   => $options['response_type'],
-            'approval_prompt' => $options['approval_prompt'],
-        ];
+        $options['client_id'] = $this->clientId;
+        $options['redirect_uri'] = $this->redirectUri;
+        $options['state'] = $this->state;
+        
+        return $options;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


shadowhand:
It is also possible to add custom parameters via the `getAuthorizationQuery` method, but this requires overloading the method in a custom provider.

I am in favor of accepting this PR.

hajekj:
Could you please lend a hand with the tests?

ramsey:
@hajekj I think a good test would be to ensure that all options passed in are included in the return value, along with the options added inside the unit. Let me know if you need any more help on this.

hajekj:
@ramsey Should I create a completely new test function or add the test into `testScopesOverloadedDuringAuthorize` [function](https://github.com/thephpleague/oauth2-client/blob/4a91ca6ab88daa6fd36a8af39adc6fdcc14a7e8b/test/src/Provider/AbstractProviderTest.php#L240)?

shadowhand:
@hajekj you should add new test method.

hajekj:
So, before I commit the test, would this be an acceptable test?
```php
/**
 * Tests https://github.com/thephpleague/oauth2-client/pull/485
 */
public function testCustomAuthorizationUrlOptions()
{
    $this->assertContains('foo=BAR', $this->provider->getAuthorizationUrl([
        'foo' => 'BAR'
    ]));
}
```

shadowhand:
That's alright... I would go about it a little more explicitly:

```php
$url = $this->provider->getAuthorizationUrl(...);
$query = parse_url($url, PHP_URL_QUERY);
$this->assertNotEmpty($query);

parse_str($query, $params);
$this->assertArrayHasKey('foo', $params);
$this->assertSame('BAR', $params['foo']);
```

hajekj:
Okay, I have added the test and the checks have passed.

shadowhand:
:+1:

ramsey:
Thanks!

PR #486, opened 2016-02-16
Fix code styling in docs

shadowhand:
As per recent changes made to Github Pages, Prism is no longer a
supported highlighter.

https://github.com/blog/2100-github-pages-now-faster-and-simpler-with-jekyll-3-0


reinink:
Thanks @shadowhand!

PR #490, opened 2016-02-24
Add Azure Active Directory provider

hajekj:
The provider is already listed on the [official site](http://oauth2-client.thephpleague.com/providers/thirdparty/) just missing here.


shadowhand:
:+1:

PR #493, opened 2016-03-04
Add documentation around authentication headers in custom provider class

garethellis36:
See #492 - add some documentation to indicate that when creating a custom `Provider` using `AbstractProvider`, it may be necessary to override the `getAuthenticationHeaders()` method.

PR #496, opened 2016-03-07
Update provider guide to refer to authentication traits

garethellis36:
Improvement on #493 to refer to the traits included in this package.

### DIFF
--- a/README.PROVIDER-GUIDE.md
+++ b/README.PROVIDER-GUIDE.md
@@ -40,22 +40,19 @@ service. If this is done via headers, you should override this method:
 protected function getAuthorizationHeaders($token = null);
 ```
 
-This method takes an `AccessToken` instance, an access token
-string or `null`, and by default returns an empty array. If the oauth2 
-service requires a `Bearer` header in the request header for an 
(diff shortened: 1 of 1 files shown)


README.PROVIDER-GUIDE.md
```diff
@@ -40,22 +40,33 @@ service. If this is done via headers, you should override this method:
 protected function getAuthorizationHeaders($token = null);
 ```
 
-This method takes an `AccessToken` instance, an access token
-string or `null`, and by default returns an empty array. If the oauth2 
-service requires a `Bearer` header in the request header for an 
-authenticated request, you could implement something like this:
-
+This package comes with a two traits for implementing standard header-based
+ authenticated requests: `Bearer` and `MAC`. To use `Bearer` authorization,
```
shadowhand: No one knows if the `MAC` trait works... none of the clients I know of use it.

garethellis36: Heh. Should I remove that reference?

shadowhand: I would.

PR #498, opened 2016-03-07
Add Betaseries provider

florentsorel:


shadowhand:
I found a couple of issues and reported them, but nothing critical. :+1: from me.

PR #516, opened 2016-04-14
Allow using a proxy for debugging.

EricHogue:
Expose the proxy and verify Guzzle options to make it possible to use a proxy. 

Verify is exposed only if a proxy is setup. Not sure if it make sense, but I don't like allowing to bypass ssl verification.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -122,7 +122,8 @@ abstract class AbstractProvider
         $this->setRequestFactory($collaborators['requestFactory']);
 
         if (empty($collaborators['httpClient'])) {
-            $client_options = ['timeout'];
+            $client_options = $this->getAllowedClientOptions($options);
+
             $collaborators['httpClient'] = new HttpClient(
                 array_intersect_key($options, array_flip($client_options))
             );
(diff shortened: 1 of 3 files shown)


src/Provider/AbstractProvider.php
```diff
@@ -136,6 +137,26 @@ public function __construct(array $options = [], array $collaborators = [])
...
+     *     Options include `clientId`, `clientSecret`, `redirectUri`, and `state`.
+     *     Individual providers may introduce more options, as needed.
+     * @return array The options to pass to the HttpClient constructor
+     */
+    protected function getAllowedClientOptions(array $options)
+    {
+        $client_options = ['timeout', 'proxy'];
+
+        // Only allow turning off ssl verification is it's for a proxy
+        if (array_key_exists('proxy', $options) && strlen($options['proxy']) > 0) {
```
shadowhand: This can be simplified to:

```php
if (!empty($options['proxy'])) { ... }
```

EricHogue: I'll do that tonight.

EricHogue: Done

shadowhand:
Thank you!

PR #518, opened 2016-04-18
Add support for capturing access token values

shadowhand:
Vendors (such as Google) sometimes pass additional information with
access token responses. Capturing these values allows package users to
make use of them when appropriate.

Refs thephpleague/oauth2-google#14

### DIFF
--- a/src/Token/AccessToken.php
+++ b/src/Token/AccessToken.php
@@ -46,6 +46,11 @@ class AccessToken implements JsonSerializable
     protected $resourceOwnerId;
 
     /**
+     * @var array
+     */
+    protected $values = [];
+
+    /**
      * Constructs an access token.
      *
      * @param array $options An array of options returned by the service provider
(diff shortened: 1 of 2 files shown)


shadowhand:
@ramsey @rtheunissen this is pretty simple and gives a lot more flexibility to users. Thoughts?

src/Token/AccessToken.php
```diff
@@ -84,6 +89,17 @@ public function __construct(array $options = [])
...
+        // Capure any additional values that might exist in the token but are
+        // not part of the standard response. Vendors will sometimes pass
+        // additional user data this way.
+        $this->values = array_diff_key($options, array_flip([
+            'access_token',
+            'resource_owner_id',
+            'refresh_token',
+            'expires_in',
+            'expires',
+        ]));
```
ramsey: Do you think it would be good for us to put our standard keys in a property or something? Ideally, this would be an array class constant, but we would need to enforce PHP >= 5.6.0 for that, and I don't think we should jump to that just yet. :-)

ramsey:
Overall, looks great to me.

src/Token/AccessToken.php
shadowhand: Maybe eventually. For now all the keys are in the same function so it seems fine to me.

shadowhand:
Released in v1.4.0.

PR #520, opened 2016-04-29
Ensure generated query strings are consistent

### LINKED ISSUE #519: http_build_query in AbstractProvider
In AbstractProvider, the arg_separator in all http_build_query should be provided to ensure it respects standard url_encoded separator required by most OAuth2 server. It should be something like: 

http_build_query($params, null, '&');

In php.ini, arg_separator.output can be redefined which may cause problem on some server. Forcing the parameter to & in http_build_query will solve this issue.

I corrected this issue in my own provider, but correcting it in AbstractProvider may avoid some debugging times to others.

shadowhand:
PHP configuration can be modified to set a different argument
separator from the standard `&`. Generated query strings should always
use the standard separator.

Fixes #519

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # OAuth 2.0 Client Changelog
 
+## 1.4.1
+
+_Released: 2016-04-29_
+
+* Add `QueryBuilderTrait` to standardize query string generation.
+
 ## 1.4.0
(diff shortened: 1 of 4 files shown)


shadowhand:
@ramsey look good?

ramsey:
Yep. Looks great.

PR #525, opened 2016-05-25
Should not throw nasty errors when expires_in is not set

pravindahal:
expires_in is "RECOMMENDED" as of RFC6749 but not required so there could be cases when neither is defined causing nasty errors on getExpires or hasExpired.

### DIFF
--- a/src/Token/AccessToken.php
+++ b/src/Token/AccessToken.php
@@ -76,6 +76,9 @@ class AccessToken implements JsonSerializable
         // We need to know when the token expires. Show preference to
         // 'expires_in' since it is defined in RFC6749 Section 5.1.
         // Defer to 'expires' if it is provided instead.
+        // Initialize 'expires' to false, meaning that it is not known
+        // or doesn't expire
+        $this->expires = false;
         if (!empty($options['expires_in'])) {
             $this->expires = time() + ((int) $options['expires_in']);
         } elseif (!empty($options['expires'])) {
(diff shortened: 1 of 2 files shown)


src/Token/AccessToken.php
```diff
@@ -76,6 +76,9 @@ public function __construct(array $options = [])
         // We need to know when the token expires. Show preference to
         // 'expires_in' since it is defined in RFC6749 Section 5.1.
         // Defer to 'expires' if it is provided instead.
+        // Initialize 'expires' to false, meaning that it is not known
+        // or doesn't expire
+        $this->expires = false;
```
shadowhand: Why `false` instead of `null`?

shadowhand:
I don't see a reason for this change.

src/Token/AccessToken.php
pravindahal: No reason in particular. I think it makes sense that expires is false means that it doesn't expire... but could also be null as in it is not defined.

pravindahal:
It is just for gracefully handling the fact that 'expires_in' is barely a recommended field as of the RFC. But feel free to reject it.

shadowhand:
But that's already handled fine. If no `expires` or `expires_in` is set, [then `getExpires` will return `null](https://github.com/thephpleague/oauth2-client/blob/master/src/Token/AccessToken.php#L139-L147). This is documented and clear.

pravindahal:
I see now. My bad.

PR #532, opened 2016-06-20
Add support of base urls containing query params

kustov-an:
A small change allowing using query params in base authorization urls

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -422,7 +422,8 @@ abstract class AbstractProvider
         $query = trim($query, '?&');
 
         if ($query) {
-            return $url . '?' . $query;
+            $glue = strstr($url, '?') === false ? '?' : '&';
+            return $url . $glue . $query;
         }
 
         return $url;
--- a/test/src/Provider/AbstractProviderTest.php
+++ b/test/src/Provider/AbstractProviderTest.php
@@ -64,7 +64,7 @@ class AbstractProviderTest extends \PHPUnit_Framework_TestCase
             'state' => 'XXX'
         ]));
     }
-    
+
     /**
      * Tests https://github.com/thephpleague/oauth2-client/pull/485
      */
@@ -75,7 +75,7 @@ class AbstractProviderTest extends \PHPUnit_Framework_TestCase
         ]);
         $query = parse_url($url, PHP_URL_QUERY);
         $this->assertNotEmpty($query);
-        
+
         parse_str($query, $params);
         $this->assertArrayHasKey('foo', $params);
         $this->assertSame('BAR', $params['foo']);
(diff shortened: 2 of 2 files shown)


src/Provider/AbstractProvider.php
```diff
@@ -422,7 +422,8 @@ protected function appendQuery($url, $query)
         $query = trim($query, '?&');
 
         if ($query) {
-            return $url . '?' . $query;
+            $glue = strstr($url, '?') === false ? '?' : '&';
+            return $url . $glue . $query;
```
shadowhand: This makes sense but raises the question as to what is supposed to happen with `$url` already contains some of the parameters in the `$query`?

shadowhand:
This PR should contain a test case.

stevenmaguire:
@kustov-an I will echo @shadowhand request for a test case. Could you please include one in this PR?

ramsey:
Thanks for providing test cases. Before I accept the PR, can you address @shadowhand's question:

> what is supposed to happen with $url already contains some of the parameters in the $query?

Maybe it's as simple as adding test cases like this:

```
['test.com/?a=1&b=2', 'test.com/?a=1&b=2', 'b=2'],
['test.com/?a=1&b=2', 'test.com/?a=1&b=1', 'b=2'],
```

And then updating the code accordingly, assuming my test cases above provide the intended/expected behavior.

j4m3s:
@ramsey, I'd like to see this PR (or another to address the issue) merged.  IMHO the answer to @shadowhand's question is: nothing special - duplicates should be appended, for two reasons:

1. The [RFC](https://tools.ietf.org/html/rfc3986#section-3.4) doesn't prohibit duplicate keys - either with the same or different values.  We should not try to be clever and infer our own rules or query spec without knowing what the caller intended. I read on stack overflow that .NET presents duplicate key values as an array - so doing something clever here could break an OAuth server's intended functionality.
2. This method is just responsible for appending the query to the url.  Anything above and beyond that is new functionality for a new method - if it can be justified based on item 1 :)

ramsey:
@j4m3s, I think that's fine, too. I'd like to see duplicates added to the test cases on this PR, so that we can ensure that our tests cover all the expected behaviors, and if the expected behavior is to duplicate parameters, then we need tests that show that. :-)

j4m3s:
You're right, the extra tests would make it clear what's expected.  I think you'll need to do that if that's ok @kustov-an?  I don't think I can modify your PR.  

Would you be able to add two test cases as follows, below the others you've added:
```
['test.com/?a=1&b=1&b=2', 'test.com/?a=1&b=1', 'b=2'],
['test.com/?a=1&b=2&b=2', 'test.com/?a=1&b=2', 'b=2'],
```

kustov-an:
Added those cases.

PR #533, opened 2016-06-29
Add Clever client to third party providers list.

smoakey:
Add Clever client to third party providers list.


shadowhand:
I opened some issues for problems I saw in the implementation. Overall the provider appears well done.

👍

Issue #475, opened 2015-12-31
Introducing HTTPlug


sagikazarmark:
Hello there,

I am one of the authors of [HTTPlug](http://httplug.io), an HTTP Client abstraction library.

We plan to use this library to implement an OAuth [authentication method](http://docs.httplug.io/en/latest/components/authentication/) for our HTTP Mesage library.

I noticed that you rely directly on Guzzle 6 for the HTTP communication. Would you be interested in using an abstraction instead?

We are planning to prepare a two-step migration process to ease migration to HTTPlug without breaking BC. (Details in this issue: https://github.com/php-http/documentation/issues/57)

While we are not stable yet, we encourage people to get known with HTTPlug, as the API is quite stable and we don't plan changing it anymore. FOS HTTP Cache migration is WIP (see this PR https://github.com/FriendsOfSymfony/FOSHttpCache/pull/257)

Let us know if you are interested or need any more information. Our documentation is available here: http://docs.httplug.io

We are also available on slack: http://slack.httplug.io

Thanks,
Mark

shadowhand:
We were using ivory before. See #315 as to when it was removed. I personally have nothing against HTTPlug but the docs are currently lacking and, as [I have mentioned before][1], not all PSR-7 implementations include all the interfaces, which makes it harder to use HTTPlug as a one-size-fits-all solution.

[1]: https://github.com/satooshi/php-coveralls/issues/80#issuecomment-165829065

sagikazarmark:
Thanks for the reply @shadowhand. I agree our documentation needs improvement, we are working on it, but there are already examples of usage.

As for the PSR-7 argument: I explained in the issue you linked above that HTTPlug does not rely directly on any PSR-7 implementations. We provide message factory interfaces which standardizes construction logic in order to make adapters/clients implementation independent.

ddeboer:
@shadowhand [The HTTPlug docs](http://php-http.readthedocs.org/en/latest/) have improved a lot over the last couple of days. Please have a look and tell us if anything is still unclear.

sagikazarmark:
@shadowhand especially this part should address your concerns:

http://php-http.readthedocs.org/en/latest/httplug/library-developers.html#messages

ramsey:
Is there a specific advantage in switching the HTTP client library that we use? I'm not opposed to changing. I just want to understand the benefits of using something else and why we should make this change.

Thanks!

sagikazarmark:
[...]

Issue #286, opened 2015-05-06
Update Guzzle to ~5.0?


rtheunissen:
I just installed the client with composer and saw this:

> guzzle/guzzle suggests installing guzzlehttp/guzzle (Guzzle 5 has moved to a new package name. The package you have installed, Guzzle 3, is deprecated.)

Any reason not to patch to the new version?

stevenmaguire:
To quote @ramsey from #214 

> The 1.0 branch is already running Guzzle 5. :-)

rtheunissen:
Aw sweet! Didn't realise there was a 1.0. Thanks. :+1:

benjy:
The 1.0 branch is still in development though right? Is 1.0 going to be compatible with the minor versions leading up to 1?

shadowhand:
@benjy 1.0 is a major version bump and does include API changes. The majority of these are internal, (most) of the external API remains the same.

benjy:
Do you have an ETA for the first tagged 1.0.x release?

shadowhand:
Hopefully very soon. Definitely within the next month.

benjy:
OK great, thanks for the info.

Issue #260, opened 2015-04-13
Move to guzzle 5


jelkevdh:
Installing this client I get the message that Guzzle 3 is deprecated.
Is it an idea to see if an up-to-date version of Guzzle can be supported?
<code><pre>composer require league/oauth2-client
Using version 0.10.* for league/oauth2-client
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
  - Installing guzzle/guzzle (v3.9.3)
    Downloading: 100%         

  - Installing league/oauth2-client (0.10.1)
    Downloading: 100%         

guzzle/guzzle suggests installing guzzlehttp/guzzle (Guzzle 5 has moved to a new package name. The package you have installed, Guzzle 3, is deprecated.)
</pre></code>

ramsey:
We are moving to Guzzle 5 for the 1.0 release. The 0.x releases will remain on Guzzle 3.

Thanks!

PR #538, opened 2016-07-19
Implement HTTPlug

Nyholm:
This is a backwards compatible implementation of [HTTPlug](http://docs.php-http.org/en/latest/httplug/introduction.html). It provides an abstraction over Guzzle and other libraries sending HTTP messages. Using HTTPlug will comply with the dependency inversion principle.

With this PR we can now allow people using Guzzle5 using this library. This will also future proof us whenever Guzzle7 is released.

### DIFF
--- a/README.UPGRADING.md
+++ b/README.UPGRADING.md
@@ -2,6 +2,6 @@
 
 ## Upgrade Guide
 
-### Upgrading to Version 1.0.0
+### Upgrading to Version 1.5
 
-TBD
+You need to install `php-http/guzzle6-adapter` and ` php-http/message`, then use the libaray as normal.
--- a/README.md
+++ b/README.md
@@ -249,8 +249,16 @@ $provider = new \League\OAuth2\Client\Provider\GenericProvider([
 Via Composer
 
 ``` bash
-$ composer require league/oauth2-client
+$ composer require league/oauth2-client php-http/guzzle6-adapter php-http/message
 ```
+
+The two extra packages from `php-http` is part of the [HTTPlug](http://httplug.io/) organisation. It helps us not to 
+be coupled to Guzzle. That is why you need to install the `php-http/guzzle6-adapter` or any other library providing the virtual 
+[php-http/client-implementation](https://packagist.org/providers/php-http/client-implementation) package. The 
+`php-http/message` package contains factory classes to create Guzzle and Diactoros PSR-7 requests. 
+
+You can read more about HTTPlug and how to use it in 
+[their documentation](http://docs.php-http.org/en/latest/httplug/users.html).
 
 ## Contributing
 
--- a/composer.json
+++ b/composer.json
@@ -4,16 +4,23 @@
     "license": "MIT",
     "require": {
         "php": ">=5.5.0",
+        "psr/http-message": "^1.0",
         "ext-curl": "*",
         "ircmaxell/random-lib": "~1.1",
-        "guzzlehttp/guzzle": "~6.0"
+        "php-http/httplug": "^1.0",
+        "php-http/client-implementation": "^1.0",
+        "php-http/message-factory": "^1.0.2",
+        "php-http/discovery": "^1.0"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.0",
         "mockery/mockery": "~0.9",
         "squizlabs/php_codesniffer": "~2.0",
         "satooshi/php-coveralls": "0.6.*",
-        "jakub-onderka/php-parallel-lint": "0.8.*"
+        "php-http/message": "^1.0",
+        "php-http/guzzle6-adapter": "^1.0",
+        "jakub-onderka/php-parallel-lint": "0.8.*",
+        "symfony/phpunit-bridge": "^3.1"
     },
     "keywords": [
         "oauth",
--- a/composer.json
+++ b/composer.json
@@ -19,8 +19,7 @@
         "satooshi/php-coveralls": "0.6.*",
         "php-http/message": "^1.0",
         "php-http/guzzle6-adapter": "^1.0",
-        "jakub-onderka/php-parallel-lint": "0.8.*",
-        "symfony/phpunit-bridge": "^3.1"
+        "jakub-onderka/php-parallel-lint": "0.8.*"
     },
     "keywords": [
         "oauth",
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -99,7 +99,8 @@ abstract class AbstractProvider
     /**
      * The options given to the Guzzle6 client
      * @var array
-     * @deprecated These are to be removed in the next major release. These options are only used in getHttpClient to keep BC.
+     * @deprecated These are to be removed in the next major release.
+     *      These options are only used in getHttpClient to keep BC.
      */
     private $guzzle6Options = [];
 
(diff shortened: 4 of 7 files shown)


shadowhand:
Nobody has ever complained about us using Guzzle directly. What is the impetus for making this PR?

Nyholm:
Thank you for the review. The main reson I made this PR is because it is a step towards better software design, better interoperability which means that more users can use this library. 

6 month ago we [made a PR](https://github.com/thephpleague/oauth2-client/issues/475) with some excellent arguments from @sagikazarmark.

> Nobody has ever complained about us using Guzzle directly.

That is not quite right... Here are issues and PRs related to you depending on Guzzle. You could expect the same wave of issues when Guzzle 7 is released. 

* https://github.com/thephpleague/oauth2-client/issues/115
* https://github.com/thephpleague/oauth2-client/pull/175
* https://github.com/thephpleague/oauth2-client/issues/215
* https://github.com/thephpleague/oauth2-client/pull/223
* https://github.com/thephpleague/oauth2-client/issues/260
* https://github.com/thephpleague/oauth2-client/issues/286
* https://github.com/thephpleague/oauth2-client/issues/314

shadowhand:
@Nyholm all of those tickets are about upgrading to the latest version of Guzzle, not replacing Guzzle with something else or requesting that we support older versions.

Guzzle 7 will most likely use PSR-7 and so very little would change for us. As far as I know, Guzzle 7 doesn't even exist right now in any form.

Nyholm:
Yes @shadowhand, that is exactly the point. Since we are dependent on Guzzle we have to deal with issues related to Guzzle. We have to be aware of changes in that dependency. If we, however, were dependent on an abstraction we would not have to deal with such issues. 

> Guzzle 7 will most likely use PSR-7 and so very little would change for us.

Yes, they would use PSR-7 and they would probably use PSR-17 which I know you are working hard on. (Thank you for that btw). But Im not concerned with how much of our code we have to change, I'm concerned with us forcing users to a specific version of Guzzle. That is bad software design (Dependency inversion principle) and our users will not be happy about it. (See the issues)

And like I said earlier:

> With this PR we can now allow people using Guzzle5 using this library. 

Whenever Guzzle7 comes our users can start using that at day 1 without us doing any change to our library.

shadowhand:
> We have to be aware of changes in that dependency.

Just as we would have to be aware of changes to HTTPlug and make a version change ~~if~~ when it changes. In my opinion, HTTPlug provides no benefits for us at this time.

Nyholm:
> Just as we would have to be aware of changes to HTTPlug and make a version change if when it changes.

The dependencies I introduce are on 4 interfaces with one method, a virtual package and the discovery lib which is a school book example of SOLID. Guzzle7 is way more complex and more likely to break BC.

>  depending on HTTPlug actually increases user friction, as the end user not only has to install league/oauth2-client but also a php-http/guzzle6-adapter

Yes, that is the point. It complies with the Single Responsibility Principle. It is not "friction" it is flexibility. 

> If Guzzle 7 came out tomorrow, then the user would have to update their composer configuration to require php-http/guzzle7-adapter and personally validate that it works correctly.

All adapters follow the Liskov substitution principle. Which means you can easily swap them out and you will have no side effects. Also, you can create your own adapter if you like. 

If an implementation misbehaves, it's a bug in the implementation, not in the abstraction.

> In my opinion, HTTPlug provides no benefits for us at this time.

Guzzle5 has 8000 daily downloads. None of those people can use version 1 of this library nor any of the great features you added after version 1. Isn't that fact alone reson enough?

It seams like you intentionally ignoring my point that an abstraction is better than an implementation and you are focusing on details. Im trying to introduce good software design patterns and principals. "Guzzle 7 will most likely use PSR-7 and so very little would change for us." is simply not an argument for not using an abstraction.

ramsey:
Thank you for your contributions. We've decided not to move away from the use of Guzzle at this time; closing this PR.

Aerendir:
Premise 1: I know the discussion is now closed, but searching for the differences between Guzzle and HTTPlug I read [this](https://www.phproundtable.com/episode/httplug-guzzle-and-apis) and came here;

Premise 2: I use Guzzle and I'm good with it and have no intention to switch

but...

This discussion is unfair in my own humble and irrelevant opinion.

Force someone to do or to use something is unfair.

You, The PHP League, are one of the best groups on the web that develops really very good software.

You have a responsibility to deliver good and quality software.
Sometimes this requires to make brave choices.

My arguments for the switching are those:

1. **An abstraction is even better than a concrete implementation**: an abstraction doesn't force anyone to use something. If I'm already using Guzzle, I'll go with it; if I'm already using Buzz, I'm go with it. I'm free to use what I want (and this is one of the best freedoms in the World);
This is not a debatable opinion; **this is a fact and a _de facto_ standard and a recognized principle**:  [The Dependency Inversion principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle) is one of the [5 SOLID principles](https://en.wikipedia.org/wiki/SOLID_(object-oriented_design)) "intended to make software designs more understandable, flexible and maintainable".
Only for this, the decision of not moving to HTTPlug is wrong, again, only in my own humble opinion.
2. **Force the use of a library over another is an unfair advantage** for the chosen one (whichever it is), and this is against the "free market" (of packages in this scenario), is against the progress, **is against the improvement of software**; Is, in some ways, [similar to what is happening in EU with Google](https://www.theguardian.com/business/2017/jun/27/google-braces-for-record-breaking-1bn-fine-from-eu).

Again, I'm an happy Guzzle's user and I have no intention to switch to something else.

But I'm also an happy and passionate developer who reads a lot of code to improve himself and see which solutions other better developers adopted.

If I can learn something I'm happy. Reading your code I learned a lot of things.

**And again, you, as one the biggest groups of PHP developers IN THE PHP WORLD - this is important: we speak about the WORLD! -, have a responsibility in developing good software and respect best practices and good and recognized design principles, embrace them and improve them when possible.**

Being stick with a concrete implementation is not a good practice and not a fair decision, **especially if an alternative exists and even more so if it was already implemented in this package**.

Excuse me for my long comment: I'm only trying to make the open source world a better place.

Thank you for your efforts in developing this amazing piece of code: I'm grateful to you.

ramsey:
Any time you use a package that has dependencies on other packages, they are forcing you to use those other packages. Switching from Guzzle to something else (even if it is an abstraction) means we're forcing our users to use that abstraction. The only way we could avoid forcing users to use any other packages is by implementing everything on our own, making this library wholly self-contained.

I do not feel our use of Guzzle in this package restricts anyone's use of this package or freedom to provide an alternative implementation. In fact, Guzzle is provided merely as a convenience. We use only the `GuzzleHttp\Psr7\Request` object in our [`RequestFactory`](https://github.com/thephpleague/oauth2-client/blob/bb02e5d0e4441d1bc97e4e92d29517b9271113f3/src/Tool/RequestFactory.php) class; others may override our `RequestFactory` to provide their own implementation. Freedom from our defaults is the primary benefit of the factory.

Soullivaneuh:
> I do not feel our use of Guzzle in this package restricts anyone's use of this package

But you are, because of the composer dependency: https://github.com/thephpleague/oauth2-client/blob/aa2e3df188f0bfd87f7880cc880e906e99923580/composer.json#L10

Currently, if I don't want to use guzzle, I'll have it installed anyway. I wan't to upgrade to Guzzle 7 or use Guzzle 5 for some reason? Well I can't, because of your requirement.

Httplug was created because projects using multiple SDK are regularly whiling onto dependencies hells.

It allows to avoid that and the implementation is quite easy even for maintainer than end users. Currently, a lot of well used library switched to php-http.

Bonus: You don't have to manage HTTP clients anymore, just use the interface and you are good to go! :+1: 

Could you please re-consider it? In any case, the guzzle requirement should be optional.

Soullivaneuh:
Also, using php-http allow us to use bench of plugins, such as the logger which is very very useful for developers.

kelunik:
@ramsey If that's true, shouldn't it depend on Guzzle's PSR-7 implementation only instead of Guzzle itself? Depending on Httplug is then indeed not helpful and just a BC break.

Aerendir:
The important point is that `oauth2-client` library MUST depend on an abstract implementation.

I don't know HTTPPlug, but I know Guzzle and I prefer its implementation - Guzzle is also used by major libraries like the AWS SDK and this is a point in its favor).

kelunik:
Indeed, the `RequestFactory` should declare the PSR-7 `RequestInterface` as return type, but there's nothing wrong the using a concrete implementation in `RequestFactory`.

Soullivaneuh:
@Aerendir I think you mistaken, HTTPPlug is an abstraction to allow using the HTTP client you want. And Guzzle is part of the current usable client.

ramsey:
This conversation comes up every year and sometimes more often than that. We've discussed it at length.

I am not personally against the use of HTTPlug from a technology standpoint. However, I feel that proponents of HTTPlug have been pushy and are browbeating the topic to the point where I do not wish to support the project by including it. The `league/oauth2-client` project is not the only place where I've observed this behavior.

I typically do not enjoy being so forward with my opinions, but in this case, I feel it is necessary to explain why we have not implemented HTTPlug into this project.

sagikazarmark:
> I feel that proponents of HTTPlug have been pushy and are browbeating the topic

I'm sorry for that 😞

teohhanhui:
HTTPlug has since become [PSR-18](https://www.php-fig.org/psr/psr-18/). Perhaps it's time to seriously think about adopting it. :smile:

Aerendir:
We should use this: https://github.com/php-fig/http-client

ramsey:
@dkarlovi See #685

dkarlovi:
I saw that issue, it pointed me to this PR which is closed, so don't really know what's next.

ramsey:
A new PR that implements this using PSR-18 is what's next. 😁

Issue #542, opened 2016-07-22
AbstractProvider::getAuthorizationParameters() overwrites 'redirect_uri'


pounard:
I have two problems with the API in its actual state, and I'd like to know if I am doing it wrong.

First one, doing this in my controller:
```php
            // Fetch the authorization URL from the provider; this returns the
            // urlAuthorize option and generates and applies any necessary parameters
            // (e.g. state).
            $authorizationUrl = $provider->getAuthorizationUrl([
                'grant_type' => 'authorization_code',
                'redirect_uri' => $this->generateUrl('some_custom_route_in_app'),
            ]);
```
will make the redirect_uri parameter being overwritten in AbstractProvider::getAuthorizationParameters() which somehow makes me sad, because I probably don't want to overwrite getAuthorizationParameters() on my provider (it would be worst if the provider was a third-party library I could use).

So I ended up thinking that I probably just should set the 'redirectUri' parameter in the $options array of the provider constructor, but my provider may run various authorization_code or token_access requests that would need my user to be redirected on different pages depending on the action I'm actually doing with OAuth2.

TL;DR:

- redirectUri in $options: global state and hardcoded, cannot override
- redirect_uri in getAuthorizationParameters(): need to override the method on the provide else it gets overwritten by default implementation

My question is: 

> Knowing that I am using a DIC based framework (Symfony) and registering the provider in the DIC, is there a way to NOT force a default redirectUri and/or to override it properly at getAuthorizationParameters() time ? Is this is bug, if yes, can I propose a PR ?

pounard:
Oh, and I did a PR https://github.com/thephpleague/oauth2-client/pull/543

shadowhand:
Many OAuth providers require a specific redirect URI when registering the application and will fail requests that use a different URI. You can certainly extend and overload to modify this behavior.

pounard:
Yes actually after a day of coding and tests, I finally got that, checking redirect_uri to be the same that the one specified to the server along the client id and secret seems to be a recommended security check per the various RFC's and changing it doesn't make sense. I seriously was doing it wrong.

Thank you for answering.

pounard:
Actually, it make sense to be able to change the redirect URI, most providers will only check for identical scheme, host and port but won't care about the relative server path and GET parameters; in my specific use case I do need to change the GET parameters.

PR #543, opened 2016-07-22
Allow business implementations set their own redirect_uri parameter

pounard:
I explained my reasons in https://github.com/thephpleague/oauth2-client/issues/542

In some cases, I need to overwrite the redirect_uri parameter depending on the context, but my provider is a singleton in a container, so I need to be able to do it at runtime.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -355,8 +355,13 @@ abstract class AbstractProvider
         // Store the state as it may need to be accessed later on.
         $this->state = $options['state'];
 
+        // Business code layer might set a different redirect_uri parameter
+        // depending on the context, leave it as-is
+        if (!isset($options['redirect_uri'])) {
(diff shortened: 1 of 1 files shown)


pounard:
Hum forget it it won't work.

pounard:
In fact, this will work as per #542

src/Provider/AbstractProvider.php
```diff
@@ -355,8 +355,13 @@ protected function getAuthorizationParameters(array $options)
         // Store the state as it may need to be accessed later on.
         $this->state = $options['state'];
 
+        // Business code layer might set a different redirect_uri parameter
+        // depending on the context, leave it as-is
+        if (!isset($options['redirect_uri'])) {
```
shadowhand: Should this be `!isset` or `empty`? Is an empty, but defined, `redirect_uri` valid?

shadowhand:
:+1:

PR #546, opened 2016-07-28
Change AbstractProvider getRandomState to only return alphanumeric states

johnnoel:
Before, this was using the default character set for `RandomLib/Generator`'s `generateString` which is the base 64 character set that includes "+" and "/". While "/" wasn't causing any problems, using "+" in a URL parameter (e.g. when the OAuth 2 server sends back the state in the query string), the "+" was getting interpreted as a space, which means when a straight string comparison to stored state was being done, it was returning false.

This changes getRandomState to use the `Generator::CHAR_ALNUM` constant as its character set which solves this problem.


shadowhand:
Makes perfect sense. I'll consider this a hotfix. Released in 1.4.2.

PR #574, opened 2016-10-16
[Travis] Test on PHP 7.1

duncan3dc:


shadowhand:
Looks like RandomLib 1.1 has some issues with PHP 7.1 that are fixed in version 1.2. I've created #575 to address this.

shadowhand:
Please rebase this branch to master so that the tests run again.

duncan3dc:
Done :+1:

shadowhand:
Looks like there are still errors when mcrypt is enabled, as per ircmaxell/RandomLib#55

GawainLynch:
@shadowhand FWIW due to PHP 7.1, in the Bolt project we've decided to replace RandomLib with polyfills for PHP 5.x as a result of the breakage/noise coming from this change, and Scott seems pretty determined to get the mcrypt stuff out a.s.a.p.

The hard(er) part for this project is the BC due to getter/setters in the abstract class, but thought it worth mentioning.

shadowhand:
Honestly, I think this good impetus for starting a version `3.x` branch that is 7+ only. That would allow us to use type hints and drop RandomLib entirely.

GawainLynch:
@shadowhand: No objections from us, but we (and a lot of projects) do need to support SF 2.8 (and 3.x from Q2/Q3 2017), which locks us at 5.5.9 for the foreseeable future.

So if needed, please see a :hand: raised from me if you _eventually_ need someone to help maintaining/backporting to whatever the legacy branch becomes.

ramsey:
What's the status of this PR? Is more development being done to attempt to fix the failures on 7.1, or should I close this for now?

shadowhand:
@ramsey it should stay open. I will have time to dig in further this week.

shadowhand:
I'm going to address this by starting a 3.x branch that makes a few changes:

- requires PHP 5.6
- switches RandomLib for random_compat
- updates to newest PHPUnit and drops Mockery

This will bring our dependencies up to latest releases and give us a solid base for starting a 4.x branch that drops PHP 5.x support.

GawainLynch:
@shadowhand could we request that PHP 5.5.9 be the base PHP version for 3.x instead? As those of us producing products based on Symfony have that as a requirement. 

At the end of the day, I don't think it would be too a huge pain for us, but it is often hard to explain upstream decisions to people and have them not respond with "Why don't you just use something else?"

shadowhand:
@GawainLynch what would be the purpose of that? [PHP 5.5 is EOL](http://php.net/eol.php) and no longer receiving patches.

If you're locked into PHP 5.5, you can continue to use the 2.x series. The 3.x series will not change functionality, it will simply drop RandomLib for `random_bytes` (via `random_compat`).

GawainLynch:
@shadowhand Supporting Symfony stack though to 4's release … but yeah, if it isn't good for you, no great stress. Just figured it couldn't hurt to ask, as RandomLib + PHP 7.1 is kinda painful for us right now too, we're in the position where we can extract it from Bolt's core use, but for our OAuth use-case it might still bite us, TBH haven't looked that hard yet as, well, RandomLib. :laughing:

Art4:
@shadowhand 
> I'm going to address this by starting a 3.x branch...

Why to start with a 3.x branch and not with 2.x?

shadowhand:
@art4 replacing RandomLib results in API changes, since we will no longer have the `getRandomFactory` and `setRandomFactory` methods. Since we follow [semver](http://semver.org/) this requires a new major version.

Art4:
I understand, but the last version is 1.4.2 (what I assume is calling the 1.x branch). Why do you want to skip the 2.x branch?

shadowhand:
Oops, for whatever reason I thought we were on 2.x already. Thanks for pointing that out!

Art4:
You're welcome :smile:

shadowhand:
Fixed by #583.

PR #578, opened 2016-10-26
Function to passing additional headers to request

Zaszczyk:
It is possible to pass additional headers to getAccessToken request. I implemented that, because I need them for TomTom OAuth2 Authorization: http://developer.tomtom.com/products/sports/mysportscloud/documentation/GetOAuth20Token.

I'll be glad for merging :)

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -521,7 +521,7 @@ abstract class AbstractProvider
     {
         $options = ['headers' => ['content-type' => 'application/x-www-form-urlencoded']];
         if (isset($params['headers']) && is_array($params['headers'])) {
-            $options['headers'] = array_merge($params['headers'], $options['headers']);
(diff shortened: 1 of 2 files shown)


mipapo:
Looks good!

Zaszczyk:
@ramsey could You look at this one?

ramsey:
@shadowhand Could you take a look at this? I seem to recall discussing something similar in the past, and I think we decided against allowing random headers to be set directly on the provider, but I don't know if I'm remember incorrectly or, if correct, why we decided that. Do you recall?

shadowhand:
This is unnecessary because providers can simply overload the `AbstractProvider::getAccessTokenOptions` to add their own headers as necessary:

```php
{
    $options = parent::getAccessTokenOptions($params);
    $options['headers']['api-key'] = 'my-api-key';

    return $options;
}
```

The OAuth spec states what is actually required, and this header is not. Thus is should be implemented on a per-provider basis, not universally.

PR #581, opened 2016-10-28
Extract method sendRequest() to new method: getRawResponse()

Zaszczyk:
I've extracted method AbstractProvider::sendRequest() from Provider\AbstractProvider::getResponse() to new method AbstractProvider::getRawResponse() because it was impossible to get status code of response. Method getResponse() returns only body of response, but some APIs return only status code, without body.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -641,12 +641,25 @@ abstract class AbstractProvider
      */
     public function getResponse(RequestInterface $request)
     {
-        $response = $this->sendRequest($request);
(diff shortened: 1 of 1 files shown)


stevenmaguire:
This smells a bit. Is there a better way to resolve your concerns without changing the API?

ramsey:
I agree with @stevenmaguire, and I'm declining this PR. If a provider needs access to the response status code, they're able to extend the AbstractProvider and override these methods.

I appreciate your contribution and your interest in this library. I hope you'll continue to use and contribute to it. Thanks!

PR #584, opened 2016-11-03
Replace Mockery with Phony

shadowhand:
The expect-run-verify model of testing does not mirror how code is
actually developed. A mock-run-verify approach allows faster prototyping
and reflects how code is actually executed.

Phony has additional features, such as order verification.

https://github.com/eloquent/phony#inception


shadowhand:
@ramsey any objections to this change?

ramsey:
@shadowhand No objections. Looks good.

PR #585, opened 2016-11-04
Updated Travis CI HHVM configuration.

ezzatron:
Here's the Travis config I was using for testing eloquent/phony#204. It adds multiple HHVM targets, including 3.6, 3.9, 3.12, and 3.15.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,36 @@
 language: php
 
-php:
-  - 5.6
-  - 7.0
-  - 7.1
-  - hhvm
+matrix:
+    include:
+        - php: 5.6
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,10 +1,36 @@
...
+          sudo: required
+          dist: trusty
+          group: edge
+        - php: hhvm-nightly
+          sudo: required
+          dist: trusty
+          group: edge
+    fast_finish: true
+    allow_failures:
+        - php: 7.1
```
shadowhand: We don't want to allow this one to fail. I will fix it post-merge.

PR #587, opened 2016-11-06
Allow fetching unparsed responses

shadowhand:
OAuth2 Client providers generally make good API gateways, due to the
inclusion of `getAuthenticatedRequest`. However, the `getResponse`
method always assumes that:

1. The response can be parsed by the provider.
2. The response headers can be discarded.

These two assumptions make it hard to use as a general purpose API
gateway in many situations where content does not come back as JSON or
when headers are required.

Thanks to @rtheunissen for most of this code, which was adapted for
changes made in version 2.x.

Fixes #408


shadowhand:
@rtheunissen how does this look?

rtheunissen:
@shadowhand sorry for being late on this, but for what it's worth, looks good to me. 👍

PR #589, opened 2016-11-08
Fix broken links

bocharsky-bw:


bocharsky-bw:
I suppose these changes will break links your website, right?
http://oauth2-client.thephpleague.com/providers/league/

shadowhand:
Right, these are links for github pages.

bocharsky-bw:
But this way all your links on GitHub is broken, that is bad too :)

PR #590, opened 2016-11-08
Updated not working links

maciekpaprocki:


shadowhand:
Duplicate of #589. Using relative links might work for both github pages and source.

PR #597, opened 2016-12-21
Updated Third Party Providers

ajquick:
Added Shopify.

https://github.com/multidimension-al/oauth2-shopify


shadowhand:
Looks good. I added a couple of issues that would improve the implementation.

PR #601, opened 2017-01-12
Remove unnecessary bootstrap file

shadowhand:
I don't see a need for this, as the tests only need to be run with composer autoloader...?


phpunit.xml
```diff
@@ -1,19 +1,18 @@
...
-		<whitelist>
+    <filter>
+        <whitelist>
             <directory suffix=".php">src/</directory>
         </whitelist>
-	</filter>
+    </filter>
 
-	<logging>
-        <log type="tap" target="build/report.tap"/>
```
shadowhand: Removed this because the `TAP` reporter is deprecated and Coveralls doesn't use it.

PR #602, opened 2017-01-12
Release version 2.x

shadowhand:
Improves compatibility with PHP 7.1, which is now officially released and has growing adopting.


shadowhand:
🎉

PR #606, opened 2017-01-19
Allow expires_in equals 0

artad:
Some services (http://vk.com, no expiration) return "expires_in" equals 0 and no "expires". 
In this case:
!empty($options['expires_in']) == FALSE 
!empty($options['expires']) == FALSE
and expires property remain NULL

### DIFF
--- a/src/Token/AccessToken.php
+++ b/src/Token/AccessToken.php
@@ -77,14 +77,11 @@ class AccessToken implements JsonSerializable
         // 'expires_in' since it is defined in RFC6749 Section 5.1.
         // Defer to 'expires' if it is provided instead.
         if (isset($options['expires_in'])) {
-
             if (!is_numeric($options['expires_in'])) {
                 throw new \InvalidArgumentException('expires_in value must be an integer');
             }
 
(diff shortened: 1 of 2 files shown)


ramsey:
Thanks, @artad.

This seems to comply with the definition of "expires_in" in [appendix A.14. of RFC 6749](https://tools.ietf.org/html/rfc6749#appendix-A.14):

    expires-in = 1*DIGIT

It can be `1` or more `DIGIT`, where [RFC 5234 defines `DIGIT`](https://tools.ietf.org/html/rfc5234#appendix-B.1) as

         DIGIT          =  %x30-39
                                ; 0-9

Any concerns, @shadowhand?

shadowhand:
LGTM.

ramsey:
Thanks!

artad:
Thanks for your attention

fjeddy:
This breaks the code tho, `hasExpired`uses the PHP `empty()` function to determine if expires_in is set or not. 0 is considered empty using that function.

As is, it yields the following error

`RuntimeException: "expires" is not set on the token in vendor/league/oauth2-client/src/Token/AccessToken.php on line 174`

PR #608, opened 2017-01-25
PHP 5.5 no longer supported

fbonzon:
composer.json says "php": ">=5.6.0"


ramsey:
Thanks for catching this!

shadowhand:
Bah, we need to make `nightly` an allowed failure.

ramsey:
Failure occurred on `php-nightly`. I'm not sure why Travis CI reported it as a failed build, since we allow failures for `php-nightly`.

Issue #609, opened 2017-01-26
Curious code construct for AbstractProvider::$state


holtkamp:
What is the use of the combination of these two lines?

- https://github.com/thephpleague/oauth2-client/blob/514c1056b879434c64263163b1986f4497791dd3/src/Provider/AbstractProvider.php#L322
- https://github.com/thephpleague/oauth2-client/blob/514c1056b879434c64263163b1986f4497791dd3/src/Provider/AbstractProvider.php#L331

Seems 'useless' as this effectively reads as:

```php
// Store the state as it may need to be accessed later on.
$this->state = $options['state'];
 ...
$options['state'] = $this->state;
```

Maybe I am missing the point though 😄

ramsey:
That is curious, indeed. It looks like this change introduced that: https://github.com/thephpleague/oauth2-client/commit/826991ec41d8d604ae70017f1988ad1bd20f0206

We were previously setting the state directly on a new array that was returned, rather than setting values on the options array passed in and returning the same array. So, line 331 can be dropped. If you'd like to submit a PR for that, we'll accept it.

Thanks!

ramsey:
Merged in #610

PR #610, opened 2017-01-26
Remove useless setting of $options['state']

holtkamp:
As pointed out in https://github.com/thephpleague/oauth2-client/issues/609#issuecomment-275448793


ramsey:
Thanks!

PR #616, opened 2017-02-02
Add provider redirect tool

### LINKED ISSUE #600: Following redirects while maintaining authorization.
I recently encountered some odd behavior with the Nest API in which the service was returning a response with `307` redirect status and a new url in the `Location` header. 

One user of the `oauth2-nest` client spotted the issue and [began researching on the Nest forums with a thread](https://nestdevelopers.io/t/unauthorized-with-seemingly-valid-access-token/556). That thread surfaced a [similar issue with the `requests` python module](https://github.com/kennethreitz/requests/issues/2949). 

In the end, Nest was sending back a redirect and Guzzle, in its default configuration, was following the redirects without the `Authorization` header and therefore returning an unauthorized response.

I have implemented a solution directly within the `oauth2-nest` client as a trait that seeks to override the `sendRequest` method in the `League\OAuth2\Client\Provider\AbstractProvider`. 

[https://github.com/stevenmaguire/oauth2-nest/blob/master/src/Tool/ProviderRedirectTrait.php](https://github.com/stevenmaguire/oauth2-nest/blob/master/src/Tool/ProviderRedirectTrait.php)

I was on the fence about implementing this solution in the client package because it only seems to cause frustration when using the library to send authenticatedRequests; it does not hinder the obtaining of access tokens, and I do draw a line between the library being an OAuth2 authorization package and not an API wrapper. My instinct was to not implement a solution and help the user understand how they may build a little bit of code themselves to make their own HTTP calls with the AccessToken obtained by the package. I decided to implement the solution in the package because it had a relatively small code footprint and removes a roadblock from the client's "Hello World" demo.

My question for the group here is simple. Do you see value in moving this solution into the core client as an optional tool for other client implementations that _may_ run into this problem?

stevenmaguire:
Resolves #600 

> I recently encountered some odd behavior with the Nest API in which the service was returning a response with `307` redirect status and a new url in the `Location` header. 
>
> One user of the `oauth2-nest` client spotted the issue and [began researching on the Nest forums with a thread](https://nestdevelopers.io/t/unauthorized-with-seemingly-valid-access-token/556). That thread surfaced a [similar issue with the `requests` python module](https://github.com/kennethreitz/requests/issues/2949). 
>
> In the end, Nest was sending back a redirect and Guzzle, in its default configuration, was following the redirects without the `Authorization` header and therefore returning an unauthorized response.
>
> I have implemented a solution directly within the `oauth2-nest` client as a trait that seeks to override the `sendRequest` method in the `League\OAuth2\Client\Provider\AbstractProvider`. 
>
> [https://github.com/stevenmaguire/oauth2-nest/blob/master/src/Tool/ProviderRedirectTrait.php](https://github.com/stevenmaguire/oauth2-nest/blob/master/src/Tool/ProviderRedirectTrait.php)
>
> I was on the fence about implementing this solution in the client package because it only seems to cause frustration when using the library to send authenticatedRequests; it does not hinder the obtaining of access tokens, and I do draw a line between the library being an OAuth2 authorization package and not an API wrapper. My instinct was to not implement a solution and help the user understand how they may build a little bit of code themselves to make their own HTTP calls with the AccessToken obtained by the package. I decided to implement the solution in the package because it had a relatively small code footprint and removes a roadblock from the client's "Hello World" demo.
>
> My question for the group here is simple. Do you see value in moving this solution into the core client as an optional tool for other client implementations that _may_ run into this problem?

### DIFF
--- a/test/src/Tool/ProviderRedirectTraitTest.php
+++ b/test/src/Tool/ProviderRedirectTraitTest.php
@@ -1,0 +1,127 @@
+<?php
+
+namespace League\OAuth2\Client\Test\Tool;
+
+use Eloquent\Phony\Phpunit\Phony;
+use GuzzleHttp\ClientInterface;
+use GuzzleHttp\Exception\BadResponseException;
+use League\OAuth2\Client\Tool\ProviderRedirectTrait;
+use PHPUnit_Framework_TestCase;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\RequestInterface;
+
+class ProviderRedirectTraitTest extends PHPUnit_Framework_TestCase
+{
+    use ProviderRedirectTrait;
+
+    public function getHttpClient()
+    {
+        return $this->httpClient;
+    }
+
+    public function setHttpClient(ClientInterface $httpClient)
+    {
+        $this->httpClient = $httpClient;
+
+        return $this;
+    }
+
+    public function testRedirectLimitDefault()
+    {
+        $this->assertEquals(2, $this->getRedirectLimit());
+    }
+
+    public function testSetRedirectLimit()
+    {
+        $redirectLimit = rand(3, 5);
+        $this->setRedirectLimit($redirectLimit);
+        $this->assertEquals($redirectLimit, $this->getRedirectLimit());
+    }
+
+    /**
+     * @expectedException InvalidArgumentException
+     **/
+    public function testSetRedirectLimitThrowsExceptionWhenNonNumericProvided()
+    {
+        $redirectLimit = 'florp';
+        $this->setRedirectLimit($redirectLimit);
+    }
+
+    /**
+     * @expectedException InvalidArgumentException
+     **/
+    public function testSetRedirectLimitThrowsExceptionWhenZeroProvided()
+    {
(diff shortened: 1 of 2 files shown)


stevenmaguire:
Any thoughts on this?

ramsey:
I don't have any problem with this being in the core package. @shadowhand, thoughts?

stevenmaguire:
hello

shadowhand:
Sorry for the atrocious delay here, this got lost in my inbox.

test/src/Tool/ProviderRedirectTraitTest.php
```diff
@@ -0,0 +1,127 @@
...
+     * @expectedException InvalidArgumentException
+     **/
+    public function testSetRedirectLimitThrowsExceptionWhenZeroProvided()
+    {
+        $redirectLimit = 0;
+        $this->setRedirectLimit($redirectLimit);
+    }
+
+    /**
+     * @expectedException InvalidArgumentException
```
shadowhand: Would be nice if these used the methods to expect exceptions instead of docblocks. Not something worth holding up the PR over.

PR #632, opened 2017-04-18
Allow depending on either v1 or v2 of random_compat

pravindahal:
Projects that use random_compat v1 are not compatible with oauth2-client. One such example is laravel 5.1 LTS. Since random_compat v1 and v2 are functionally same, it should allow composer decide which version to install based on what other projects use.


shadowhand:
:+1:

PR #633, opened 2017-05-17
Add yelp to third party providers

stevenmaguire:
Yelp is pushing the [v3 flavor of their API, "Fusion"](https://www.yelp.com/developers/documentation/v3/authentication). It uses OAuth2 and currently only supports Client Credentials Grant. Here is a package ripe and ready to go!


shadowhand:
Don't you have enough packages? 😉

stevenmaguire:
@shadowhand I do have enough! I also just keep finding APIs that I need to connect to via OAuth without package support. It super easy to implement, so I do it! I am currently adding v3 support to my yelp-php package and they use OAuth2 Client Credentials only!

On that topic, this was my first oauth2 package where authorization grant was not available to implement. As a result, there is also no resource owner. I took the approach of throwing exceptions if that flow is initiated. What are your thoughts on that? Would you go in a different direction?

shadowhand:
@stevenmaguire 🕚 seems fine to me.

PR #636, opened 2017-06-16
Fix TypeError thrown because getResourceOwner receives a non-JSON Response

steverhoades:
The code as written trusts that the OAuth2 server will always return a JSON response and will result in a 500 Exception by the consuming application if it receives any other response.

This fixes that issue but checking the response is what is expected before attemping to call createResourceOwner.

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -766,7 +766,15 @@ abstract class AbstractProvider
 
         $request = $this->getAuthenticatedRequest(self::METHOD_GET, $url, $token);
 
-        return $this->getParsedResponse($request);
+        $response = $this->getParsedResponse($request);
+
+        if (false === is_array($response)) {
+            throw new UnexpectedValueException(
+                'Invalid response received from Authorization Server. Expected JSON.'
+            );
+        }
+
+        return $response;
     }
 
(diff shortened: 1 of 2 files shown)


src/Provider/AbstractProvider.php
```diff
@@ -766,7 +766,15 @@ protected function fetchResourceOwnerDetails(AccessToken $token)
 
         $request = $this->getAuthenticatedRequest(self::METHOD_GET, $url, $token);
 
-        return $this->getParsedResponse($request);
+        $response = $this->getParsedResponse($request);
+
+        if (false === is_array($response)) {
+            throw new UnexpectedValueException(
+                'Invalid response received from Authorization Server. Expected JSON.'
```
shadowhand: This message could be improved, something like:

> Expected to receive a JSON response from Authorization Server

steverhoades: I agree to that.  In fact, we do have the URL that was being accessed in that block.  Instead of Authorization server we could use this value.

steverhoades:
@shadowhand any idea when we'll get a new release that includes this change?

ramsey:
Sorry for the long delay. This has been released in version 2.3.0.

PR #638, opened 2017-07-03
Updates broken link in readme

dustinalan:
Should be: https://github.com/thephpleague/oauth2-client/blob/master/docs/providers/league.md


shadowhand:
This isn't correct. The docs are deployed to http://oauth2-client.thephpleague.com/providers/league/, which removes the `docs/` from the path.

dustinalan:
Ahh, didn't visit the deployed site.

PR #640, opened 2017-07-13
Add provider for Dropbox v2 API and deprecate v1

iget-esoares:
The great @stevenmaguire's provider is based on Dropbox API v1 that's deprecated and will be shut down on 28 September.

I've create a new provider that's compatible with Dropbox API v2 (and with v1 while it's alive), 100% tested and covered as the original one.

I suggest that the documentation is updated to include my alternative


stevenmaguire:
Why not contribute to the one that already exists?

iget-esoares:
@stevenmaguire No problem, I thought that was better to keep the two versions running, but after I was writing the PR I realized that the v2 authentication is compatible with v1, so it should not be breaking changes.

Tomorrow I make a PR to the existing, I appreciate if you can merge it into a new version.

shadowhand:
Will be resolved downstream.

stevenmaguire:
Thanks @iget-esoares and @shadowhand. A PR on the original client has been created and will soon be merged.

PR #646, opened 2017-08-16
Update DigitalOcean package name

joecwallace:
If one were to copy/paste the DigitalOcean provider package name into a `composer require`, she would get a "Could not find package" error. The real package name is `chrishemmings/oauth2-digitalocean` [as seen on Packagist](https://packagist.org/packages/chrishemmings/oauth2-digitalocean).


ramsey:
Thanks, Joe!

PR #660, opened 2017-09-28
Fix the broken "first party providers" link

anaspk:
Just a simple change in the href attribute of a link so that it works on GitHub.


ramsey:
The link is correct. Please see: #638, #679, #671, and #662.

klenium:
Meybe just change the link to `http://oauth2-client.thephpleague.com/providers/league/`?

PR #662, opened 2017-10-15
Fixed link to offical providers

tempcke:
No code was changed, just the link from third party providers md file to the official providers md file.

### DIFF
--- a/docs/providers/thirdparty.md
+++ b/docs/providers/thirdparty.md
@@ -9,7 +9,7 @@ Third-Party Providers
 
 These providers allow integration with other providers not supported by `oauth2-client`. They may require an older
 version so please help them out with a pull request if you notice this. If you're looking for an official `league` provider,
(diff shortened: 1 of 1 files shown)


stevenmaguire:
@tempcke please take a look at https://github.com/thephpleague/oauth2-client/pull/638

tempcke:
@stevenmaguire does that mean you want the link to be http://oauth2-client.thephpleague.com/providers/league/ ?  If this page has the maintained list then the .md file I linked to should be removed and the link should be updated to http://oauth2-client.thephpleague.com/providers/league/ .. Do you want me to do that? seems silly when it would only take you a second to do it but I can.  Either way it should not remain the way it is now because it is broken.

stevenmaguire:
@tempcke it means: 

> This isn't correct. The docs are deployed to http://oauth2-client.thephpleague.com/providers/league/, which removes the docs/ from the path.

The path is correct.

PR #653, opened 2017-08-29
Add http basic auth

mgalagan-sugarcrm:
Add HTTP basic auth to access token request.

According to [RFC](https://tools.ietf.org/html/rfc6749#section-2.3.1) :

> Including the client credentials in the request-body using the two
   parameters is NOT RECOMMENDED and SHOULD be limited to clients unable
   to directly utilize the HTTP Basic authentication scheme (or other
   password-based HTTP authentication schemes).  The parameters can only
   be transmitted in the request-body and MUST NOT be included in the
   request URI.

PR #664, opened 2017-10-18
Support HTTP Basic Auth for client authentication

### LINKED ISSUE #576: Support HTTP Basic Auth for client authentication
When dealing with an OAuth-Server which is not supporting the authorization with credentials inside the request body the Abstract- or GenericProvider is not directly working.

So you have to implement a new provider which is overriding the getAccessTokenOptions method.

It would be better if the AbstractProvider would stick to the recommended behavior from [RFC 6749](https://tools.ietf.org/html/rfc6749#section-2.3.1) of sending the client credentials using the basic auth (authorization header). As you can see this section also tells that sending credentials inside the body is not recommended and should be limited to clients unable to use basic auth.

The getAccessTokenOptions could look like this:

``` php
    /**
     * Builds request options used for requesting an access token.
     *
     * @param  array $params
     * @return array
     */
    {
        $options = [
            'headers' => [
                'content-type' => 'application/x-www-form-urlencoded',
                'authorization' => 'Basic ' . base64_encode($params['client_id'] . ':' . $params['client_secret'])
            ]
        ];
        }

        return $options;
    }
```

As the client_id and client_secret is still inside the params array, they are also still inside the request body, so they could be removed from the params if necessary.

jenschude:
Closes #576

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -91,7 +91,7 @@ abstract class AbstractProvider
      */
     protected $httpClient;
 
-    private $authType = self::AUTH_FORM;
+    protected $authType = self::AUTH_FORM;
 
     /**
      * Constructs an OAuth 2.0 service provider.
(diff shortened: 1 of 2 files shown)


ramsey:
Thank you for contributing! 🎉

Please fix the breaking build so that we can consider this feature.

ramsey:
Duplicate of #653.

jenschude:
Revoked in favor of #653

PR #669, opened 2017-11-08
Use PHPUnit\Framework\TestCase instead of PHPUnit_Framework_TestCase

carusogabriel:
I use the `PHPUnit\Framework\TestCase` notation instead of `PHPUnit_Framework_TestCase` while extending our TestCases. This will help us migrate to PHPUnit 6, that [no longer support snake case class names](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).


ramsey:
Thank you for contributing.

Unfortunately, I made this change in eb25be9e9dface2016ebb09ff4d47e264db1c401 before looking at the open pull requests. I'll be sure not to do that in the future so that I don't steal the credit from others. :smile:

PR #671, opened 2017-11-22
 Updated broken documentation links

morozgrafix:


morozgrafix:
Didn't realize that docs were hosted elsewhere. #638 has explanation

PR #674, opened 2017-12-06
Refactoring tests

carusogabriel:
I've refactored some tests using:
- `assertArrayHasKey` instead of `isset` function;
- `assertInternalType` instead of `is_*` functions.


ramsey:
Thank you for your contribution.

PR #676, opened 2017-12-22
Avoid 'Fatal error: Wrong parameters for Exception' in error response

pradtke:
Several OAuth2 servers respond with non-standard errors, and the GenericProvider should
be robust in handling them. e.g. I shouldn't get a Fatal error.


ramsey:
Thank you for contributing! 🎉

PR #679, opened 2017-12-26
Fixed broken offical providers link

klenium:


stevenmaguire:
> This isn't correct. The docs are deployed to http://oauth2-client.thephpleague.com/providers/league/, which removes the docs/ from the path.

#638

klenium:
That's why web links should never be relative.

PR #682, opened 2018-01-13
Remove official project support for HHVM

ramsey:
This PR removes all HHVM testing on Travis CI and removes mention of official support for HHVM from the README.

While our test suite continues to pass on all HHVM versions we test, many other projects have moved away from official support for HHVM, since HHVM no longer maintains feature parity with PHP 7. When league/oauth2-client no longer supports PHP 5.6, we will need to remove support for HHVM anyway, since we'll potentially be using PHP 7+ features that HHVM does not support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,36 +1,18 @@
 language: php
 
+php:
+  - 5.6
+  - 7.0
+  - 7.1
+  - 7.2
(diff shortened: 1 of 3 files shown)


frankdejonge:
I've dropped active support for it on all my packages.

PR #690, opened 2018-02-19
Static analysis with PHPStan

lookyman:
Note that I had to remove support for PHP 5.6 and HHVM. If that is a problem, I could maybe work around that somehow (like install PHPStan only on Travis on PHP 7+), but this was an easiest solution.


ramsey:
I've begun working on a 3.x branch that sets the minimum PHP requirement to 7.1. I tried to pull this into that branch, using PHPStan 0.10, but I encountered fatal errors with the LiberatorExtension. Since I'm not familiar with how extensions work with PHPStan, I decided to back out of those changes and ask for your help. 😄 

Are you able to update your PR to target the 3.x branch and use the latest versions of the PHPStan packages?

Thanks!

lookyman:
Sure, I'll give it a go.

PR #696, opened 2018-03-08
Add support to blacklist mass assignment of specific provider properties

stevenmaguire:
Recently, I was working on implementing some new functionality for the LinkedIn provider project and ran into the need to prevent a specific property from being set on provider instantiation. The property needed to be able to support user defined values but I wanted to control that through some getters and setters. Because the current version of the AbstractProvider loops over all inbound options and sets them, if found, the constructor was completely short circuiting my goals. I wanted a more elegant way to skip specific properties on instantiation of downstream providers, so I pulled this branch together. 

It will have no impact on BC, as it defaults to all defined properties being allowed through. It will however now provide a means for a downstream provider to override and define the "guarded" properties, by name, if needed. 

```php
class ProviderWithGuardedProperties extends AbstractProvider
{
    /**
     * The properties that aren't mass assignable.
     *
     * @var array
     */
    protected $guarded = ['skipMeDuringMassAssignment'];

    /**
     * Throwaway property that shouldn't be mass assigned.
     *
     * @var string
     */
    protected $skipMeDuringMassAssignment = 'foo';

    //
}
```

I also added some tests to cover this new functionality. 

The PR also includes a couple of testing cleanup items, one specifically that brings our code coverage back up to 100%.

Anywho, not married to the idea just wanted to propose it here. Let me know what you think.

### DIFF
--- a/test/src/Provider/AbstractProviderTest.php
+++ b/test/src/Provider/AbstractProviderTest.php
@@ -2,6 +2,7 @@
 
 namespace League\OAuth2\Client\Test\Provider;
 
+use UnexpectedValueException;
 use Eloquent\Liberator\Liberator;
 use Eloquent\Phony\Phpunit\Phony;
 use GuzzleHttp\Exception\BadResponseException;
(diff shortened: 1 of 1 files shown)


stevenmaguire:
@ramsey @shadowhand Any thoughts or feelings on this one?

stevenmaguire:
what do you think

stevenmaguire:
@ramsey I noticed some housekeeping on open PRs. What do you think about this one? I've got duplicate code in two providers addressing this use case and it would be nice to strip it out :)

ramsey:
Will review this weekend. Sorry for the delay.

ramsey:
@stevenmaguire Have you considered a whitelist approach? I could go either way, but I was curious if you had considered an array to define the names of allowed properties rather than disallowed (guarded) properties.

Other than that, this looks good to me. Let me know your thoughts on the whitelist approach, though.

stevenmaguire:
@ramsey I did consider the inverse of this design but chose this approach in favor of a smoother rollout and less chance of a dot release causing the downstream providers to break, as 100% of them would require at least one new commit and release to function. Right?

PR #697, opened 2018-03-08
Add Pipedrive in thirdparty.md

daniti:


ramsey:
Thank you for contributing! 🎉

I recommend adding some tests to your library. :smile:

PR #700, opened 2018-04-15
change package

michaelKaefer:
change lemonstand/oauth2-amazon to rewritten michaelkaefer/oauth2-amazon (since lemonstand/oauth2-amazon obviously does not work as expected)


ramsey:
Thank you for contributing! 🎉

I see that the lemonstand package is using a very early version of league/oauth2-client, so making this change seems like the right thing to do.

PR #711, opened 2018-05-22
Add oauth2-pinterest to third party providers list

vantezzen:


ramsey:
Thank you for contributing! 🎉

PR #720, opened 2018-06-21
Add @throws tag for better hint when calling getAccessToken()

lx1036:
When I try...catch $this->provider->getAccessToken(), the catched exception IdentityProviderException is hinted "no throw" by the IDE. So, it should be add @throws tag to fix this issue.


ramsey:
Thank you for contributing! 🎉

PR #721, opened 2018-06-21
Add an AccessTokenInterface to allow AccessToken override

nicolassing:
I'm currently working on an implementation of openID Connect based on league/oauth2-client and I need to override the AccessToken class. 

I think it might be useful in other cases to be able to override this class. 

Anyway, thx all for your awesome work :)


src/Token/AccessTokenInterface.php
```diff
@@ -0,0 +1,79 @@
...
+     * @return integer|null
+     */
+    public function getExpires();
+
+    /**
+     * Returns the resource owner identifier, if defined.
+     *
+     * @return string|null
+     */
+    public function getResourceOwnerId();
```
ramsey: I wonder if `getResourceOwnerId()` should be removed from the base `AccessTokenInterface`. It's technically not a function of OAuth 2. Maybe we move it to a `ResourceOwnerAccessTokenInterface` that implements `AccessTokenInterface`, and then the base `AccessToken` in this library implements `ResourceOwnerAccessTokenInterface`.

ramsey:
Thank you for contributing! 🎉

PR #726, opened 2018-07-13
Throw exception on Non-JSON response from access token request

schmengler:
Otherwise a type error is thrown in the next line. The exception is the same as in `fetchResourceOwnerDetails()`


ramsey:
Thank you for contributing! 🎉

PR #741, opened 2018-10-24
Remove FreeAgent provider doc

dcorking:
as no longer maintained and probably not compatible with current oauth2-client

comment from original author: https://github.com/CloudManaged/oauth2-freeagent/issues/6#issuecomment-432433512

@ramsey do you think we should delete FA from the page all together, or do as @zot24 suggested and merely replace the maintainer name with the word 'unmaintained' in case someone chooses to fork and update it? At CloudManaged we are easy either way - we aren't taking down the repo.


dcorking:
@ramsey will this commit automatically update the web page at http://oauth2-client.thephpleague.com/providers/thirdparty/ ?

ramsey:
@dcorking Yes, this commit will automatically update that page.

> do you think we should delete FA from the page all together, or do as @zot24 suggested and merely replace the maintainer name with the word 'unmaintained'

We've never dealt with this issue that I can recall, so we don't have a precedent. I'm inclined to say that we should leave it in the list and use "Unmaintained" in place of the author name, as you suggest—perhaps italicized for emphasis.

dcorking:
@ramsey 👆 PR changed to keep FA but listed as unmaintained. I am happy to rebase if you prefer.

ramsey:
Thanks! I can do the squashing when merging. Cheers!

dcorking:
for the record
image

PR #746, opened 2018-11-10
Add new Spotify OAuth 2.0 Provider to the 3rd party providers list

ker0x:
This new provider is fully compliant with League's OAuth 2.0 Client v2


ramsey:
Thanks for the PR.

Before removing a link to someone else's project, I'd prefer to give them the opportunity to comment and approve. IMO, it's okay to have multiple libraries listed for the same provider, since each library might provide different functionality.

@jonjomckay, this concerns your [spotify-web-api library](https://github.com/jonjomckay/spotify-web-api). Is that project still active? Should we leave the link in place here?

Thanks!

jonjomckay:
I wouldn't call my library active, but as far as I'm aware, it still works!

If doesn't meet your activity requirements, I'm totally OK for it to be removed from the list though.

ramsey:
We don’t have any activity requirements. I just wanted to make sure the library is still maintained. Since it is, and since some may still use older versions of oauth2-client, I’d like to keep it in the list.

@ker0x, would you mind updating your PR to keep the link to @jonnomckay’s Spotify library, while adding yours?

Thanks!

ker0x:
Hello @ramsey,

Thanks for your feedback, I just updated my PR !

ramsey:
Thanks!

PR #747, opened 2018-11-11
Add Atlassian Jira as third-party provider

mrjoops:
Atlassian is slowly deprecating OAuth 1.0a for application development (https://developer.atlassian.com/cloud/jira/platform/oauth-2-authorization-code-grants-3lo-for-apps/), so here is a provider using OAuth 2 and the new workflow.
See https://github.com/mrjoops/oauth2-jira


ramsey:
Thank you for contributing! 🎉

PR #748, opened 2018-11-14
Add mailXpert to the list of thirdparty providers

scuben:
Adding a provider for https://dev.mailxpert.ch


ramsey:
Thank you for contributing! 🎉

PR #749, opened 2018-11-15
Fix link to first party providers

andrea-cristaudo:


ramsey:
Thanks for the PR. This link is correct when viewed from here: <http://oauth2-client.thephpleague.com/providers/thirdparty/>, so no change is necessary. Cheers!

andrea-cristaudo:
Thanks for reviewing it.
Just to let you know: the link is not correct when viewed from github web interface (https://github.com/thephpleague/oauth2-client/blob/master/docs/providers/thirdparty.md) so maybe there's something wrong in the way the html is created.

ramsey:
It’s an issue with how GitHub generates the repository sites. We have a long and storied history with this. 😀

https://github.com/thephpleague/oauth2-client/pull/638

PR #755, opened 2018-11-28
Static analysis with PHPStan

lookyman:
Replaces #690, targets 3.x branch.

So far I have been able to get to level 3. Higher levels are proving to be nontrivial. You can merge it as is, and I can continue in another PRs, or I can do some more work in this one.


ramsey:
Thank you for contributing! 🎉

PR #757, opened 2019-01-12
Fixed first party providers link

bggardner:


ramsey:
This is not a bug. Please see #638.

Issue #770, opened 2019-02-28
Broken link "There are a large number of community packages that interface with other systems."


fmertins:
Hi there, on https://github.com/thephpleague/oauth2-client/blob/master/docs/providers/league.md there is a broken link (https://github.com/thephpleague/oauth2-client/blob/master/providers/thirdparty)

sandacn:
see pull request https://github.com/thephpleague/oauth2-client/pull/772
but there is another doc site: https://oauth2-client.thephpleague.com/providers/league/

In that page, the link is not broken.

shadowhand:
The `docs/` directory is what powers the website https://oauth2-client.thephpleague.com/ Modifying links to work in the Github file browser will break the website.

fmertins:
Hi folks, I didn't know that other site, I'll have a look, awesome, thanks!

PR #772, opened 2019-03-02
Fix league link

sandacn:
refer #770 
but in this page: https://oauth2-client.thephpleague.com/providers/league/

the link is right, will the doc generator will replace the link automatically ? 
if so , my modification is right, or else, it's useless.


shadowhand:
See comment on the issue page.

Issue #691, opened 2018-02-20
Dead link on thirdparty.md


westerjn:
Low priority, but the link to the First Providers is dead for me (aka 404)

https://github.com/thephpleague/oauth2-client/blob/master/docs/providers/thirdparty.md

westerjn:
Found the first providers list at http://oauth2-client.thephpleague.com/providers/league/

Will try to edit page via PR.

ramsey:
Please see #670, #660, #638, #679, #671, and #662. This works as expected. No PR necessary. :smile:

Issue #670, opened 2017-11-20
Missing official providers list


wesamly:
If you check [Third-Party Providers](https://github.com/thephpleague/oauth2-client/blob/master/docs/providers/thirdparty.md) page, it provides [first party providers](https://github.com/thephpleague/oauth2-client/blob/master/providers/league) page which is leads to Page not found.

I was searching for Google provider, first I thought it was not available until I searched.

Thank you.

wesamly:
For now you can search https://github.com/thephpleague for oauth2- and it will list all repositories starting with oauth2- like [oauth2-google](https://github.com/thephpleague/oauth2-google)

stevenmaguire:
These markdown files are used to generate the documentation site at http://oauth2-client.thephpleague.com

The links in question function appropriately on that site...

wesamly:
@stevenmaguire I see, I checked the website and yes the link works there. Thank you.

PR #773, opened 2019-03-05
Add issue and pull request template to attempt to curb redundant reports

stevenmaguire:
I noticed that for what feels like a long time we are constantly closing redundant issues related to the allegedly broken documentation links (#670, #660, #638, #679, #671, #662, #691, #770, #792, #793).

This PR is an attempt to proactively prevent the creation of new Issues and Pull Requests by utilizing the [templating feature on github](https://help.github.com/en/articles/about-issue-and-pull-request-templates). 

I think there are a couple more things to consider...

- Should we include a similar note in `CONTRIBUTING.md`
- If we're going to introduce Issue and Pull Request templates, should we also include helpful content to actually help would be contributors compose great Issues and Pull Request descriptions? (we can do that later I suppose)

What do we all think?


stevenmaguire:
old but still lit

PR #779, opened 2019-04-12
Update thirdparty.md

matthewtrask:
fix the link from the third party page to first party page.


ramsey:
The links work properly from https://oauth2-client.thephpleague.com/. Unfortunately, we can’t make them work in both places.

PR #781, opened 2019-04-15
Auto change the return function by detecting the php version you are …

kiranbhattarai:
php version 5.x does not support funtion random_bytes(), so the script will check the php version of the server and switch between random_bytes() and mt_rand()

### DIFF
--- a/src/Provider/AbstractProvider.php
+++ b/src/Provider/AbstractProvider.php
@@ -302,7 +302,13 @@ abstract class AbstractProvider
     {
         // Converting bytes to hex will always double length. Hence, we can reduce
         // the amount of bytes by half to produce the correct length.
-        return bin2hex(random_bytes($length / 2));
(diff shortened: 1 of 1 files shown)


adbario:
There's no need for this, as [random_combat polyfill](https://github.com/paragonie/random_compat) is already included. See https://github.com/thephpleague/oauth2-client/blob/master/composer.json#L11

kiranbhattarai:
> There's no need for this, as [random_combat polyfill](https://github.com/paragonie/random_compat) is already included. See https://github.com/thephpleague/oauth2-client/blob/master/composer.json#L11

https://github.com/thephpleague/oauth2-linkedin/blob/master/composer.json Do you think i should include that over here too.

adbario:
No need to, as `thephpleague/oauth2-client` is dependency which requires the polyfill, so it's there already.

PR #788, opened 2019-05-31
Use ArrayAccessor in GenericResourceOwner

AndreiIgna:
Allow the `resourceOwnerId` to be specified in more formats, like `user.id`.
This would enable the `GenericProvider` to be used with more OAuth APIs.

Slack, for example, returns the user info in a nested array https://api.slack.com/methods/users.identity#response

By accepting this PR, the Slack OAuth and others can be used with the `GenericProvider`.

### DIFF
--- a/src/Provider/GenericResourceOwner.php
+++ b/src/Provider/GenericResourceOwner.php
@@ -14,11 +14,15 @@
 
 namespace League\OAuth2\Client\Provider;
 
+use League\OAuth2\Client\Tool\ArrayAccessorTrait;
+
 /**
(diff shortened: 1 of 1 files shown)


AndreiIgna:
Second commit adds `GenericResourceOwner::getAttribute()` helper method, which makes it easier to access user data.
```php
// at the moment
$userData = $resourceOwner->toArray();
$userField = $userData['field'];
$userField2 = $userData['group']['field'];

// with getAttribute() method
$userField = $resourceOwner->getAttribute('field');
$userField2 = $resourceOwner->getAttribute('group.field');
```

PR #789, opened 2019-06-02
Add Rocket Beans TV to 3rd Party Providers List

oliverschloebe:


bahuma20:
I've tested the provider, and it works for me. So a +1 for this.

PR #790, opened 2019-06-02
Add Twitch.tv (New API Helix) to 3rd Part Providers

vertisan:
I know, we have already provider for Twitch.tv but this is using [New API](https://dev.twitch.tv/docs/api/) known as **Helix**


Fasani:
Any reason to merge this? I will test it today also.

PR #805, opened 2019-10-09
Fix broken links in the documentation folder

cleacos:


ramsey:
Please see #670, #660, #638, #679, #671, #662, and #691. See also #773. This works as expected. No PR necessary. :smile:

PR #807, opened 2019-10-16
Added Netlify Provider

putyourlightson:
Added a Netlify Provider that works with the [Netlify API](https://docs.netlify.com/api/get-started/) to the list of third-party providers.


ramsey:
@putyourlightson @bencroker, thank you for contributing, and I'm sorry it's taken so long to respond. While we don't fully review the code of all third-party providers, we do try to make sure each provider includes tests. Since your package doesn't include tests, I'll close this PR. Full free to re-submit after you update your package with tests. Cheers!

bencroker:
Fair enough @ramsey, I'll add tests to the package and resubmit.

PR #808, opened 2019-10-18
Added Sign in with Apple third party provider

patrickbussmann:
See https://developer.apple.com/sign-in-with-apple/.


AndreiIgna:
The League OAuth2 Client supports PHP >= 5.6, but this provider does not. Saw the dependency https://github.com/web-token/jwt-framework/ requires at least 7.2

Unfortunately there's still websites using PHP lower than 7.2. I'm using OAuth 2 Client and providers in a WordPress plugin, and these sites are really slow to update their servers.. Is there a way to make Apple Sign in provider be compatible with PHP 5.6?

patrickbussmann:
I already made it 7.1 compatible, @AndreiIgna 😅

The main problem here is the JWT Generation which we need for Apple's client_secret.
Apple use's a `Elliptic Curve Digital Signature Algorithm (ECDSA) with the P-256 curve and the SHA-256 hash algorithm`.
There are many users which have `invalid_client` problems because they not generating this client_secret in the right way.
As you can read here: https://forums.developer.apple.com/community/app-frameworks/sign-in-with-apple
For example (NodeJS): https://forums.developer.apple.com/thread/125074
Or (PHP): https://forums.developer.apple.com/message/381848

I can check if I can build a own implementation of this generation and hope that it's usable for PHP 5.6. But for now this library is the only one I find.
I already checked few others but none works.

Similar problems:
https://stackoverflow.com/q/52214433/1362858

I'll try this soon: https://www.zenstruck.com/projects/php-jwt
Maybe it's working. Then we are compatible with 5.6.

Let's hope.

patrickbussmann:
OK @AndreiIgna I was successfull.
Travis is building 0.2.0 with PHP 5.6 successfully. 🎉

So it's now fully compatible with PHP 5.6.

williamxsp:
Hi!
Any news on this being merged?

ramsey:
Library quality looks good, though I do have a couple of notes:

* There's no need to use `League` as the root namespace. In fact, it's best to use your own, in case there were ever conflicts in the future, like if The PHP League decided to provide its own `League\OAuth2\Client\Provider\Apple` (we won't, but this is an example).
* Just because league/oauth2-client still supports PHP 5.6, there's no requirement for downstream providers to support older versions of PHP. It's great that you were able to get it to work for users who can't upgrade, though.

PR #813, opened 2020-01-16
Add Dokeop provider

dokeop:
Hi

Thanks for your fantastic client !
I wrote a provider for my service named [Dokeop](https://www.dokeop.com)


ramsey:
Just a tiny note:

There's no need to use `League` as the root namespace for your package. In fact, it's best to use your own, in case there were ever conflicts in the future, like if The PHP League decided to provide its own `League\OAuth2\Client\Provider\Dokeop` (we won't, but this is an example).

dokeop:
I agree for the namespace, I've just made a release with my root namespace.

PR #815, opened 2020-01-27
Add Xero provider to table

calcinai:


ramsey:
@calcinai, thank you for contributing, and I'm sorry it's taken so long to respond. While we don't fully review the code of all third-party providers, we do try to make sure each provider includes tests. Since your package doesn't include tests, I'll close this PR. Feel free to re-submit after you update your package with tests. Cheers!

PR #818, opened 2020-02-20
Don't include class in interface if not used

hendeavors:


ramsey:
Thank you for the PR! However, this is a stylistic issue that doesn't affect performance or how the code functions. In fact, PHP doesn't "import" these classes at all. The `use` statements simply create aliases.

Since some coding standards are starting to specify referencing class names by `use` statements only (e.g. [SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly](https://github.com/slevomat/coding-standard)), and we haven't adopted a specific stance on this in league/oauth2-client, I'll close this PR for now.

Feel free to open an issue to make a case for adopting a specific style regarding `use` statements.

Cheers!

PR #823, opened 2020-03-04
add more detail on the unit of time for the access token expiration

hendeavors:
I constantly find myself visiting the rfc to determine if its seconds or milliseconds.


hendeavors:
wrong branch. closing.

Issue #626, opened 2017-02-28
`AbstractProvider::getParsedResponse` should always return array or throw exception


ksimka:
```
Uncaught Exception "TypeError" with message: "Argument 1 passed to League\OAuth2\Client\Provider\AbstractProvider::prepareAccessTokenResponse() must be of the type array, string given
```

It happens because `AbstractProvider::getParsedResponse` returns string when content can't be parsed.

1. https://github.com/thephpleague/oauth2-client/blob/master/src/Provider/AbstractProvider.php#L528

```php
// AbstractProvider::getAccessToken

// ...
$response = $this->getParsedResponse($request);
$prepared = $this->prepareAccessTokenResponse($response);
// ...
```

2. https://github.com/thephpleague/oauth2-client/blob/master/src/Provider/AbstractProvider.php#L597

```php
/**
     * Sends a request and returns the parsed response.
     *
     * @param  RequestInterface $request
     * @return mixed
     */
    public function getParsedResponse(RequestInterface $request)
    {
        try {
            $response = $this->getResponse($request);
        } catch (BadResponseException $e) {
            $response = $e->getResponse();
        }
        $parsed = $this->parseResponse($response);
        $this->checkResponse($response, $parsed);
        return $parsed;
    }
```

Attention to docs — "return mixed". It can not return "mixed" (though actually it is), because the result goes to `prepareAccessTokenResponse` which receives array only.

3. https://github.com/thephpleague/oauth2-client/blob/master/src/Provider/AbstractProvider.php#L700

```php
/**
     * Prepares an parsed access token response for a grant.
     *
     * Custom mapping of expiration, etc should be done here. Always call the
     * parent method when overloading this method.
     *
     * @param  mixed $result
     * @return array
     */
    protected function prepareAccessTokenResponse(array $result)
```

Docs say `param mixed $result` (was changed here https://github.com/thephpleague/oauth2-client/commit/365e61c6b1a092f79d81c723680852213dbdedac with no visible reason), but typehint says `array $result`.

And the last thing — where mixed comes from?

4. https://github.com/thephpleague/oauth2-client/blob/master/src/Provider/AbstractProvider.php#L657

```php
/**
     * Parses the response according to its content-type header.
     *
     * @throws UnexpectedValueException
     * @param  ResponseInterface $response
     * @return array
     */
    protected function parseResponse(ResponseInterface $response)
    {
        $content = (string) $response->getBody();
        $type = $this->getContentType($response);
[...]

PR #825, opened 2020-03-13
Introduce ResponseParsingException and throw this exception in …

tandhika:
…AbstractProvider::parseResponse() if necessary


tandhika:
This PR aims at resolving issue https://github.com/thephpleague/oauth2-client/issues/626

tandhika:
The failed test was due to a bug in json_decode of php5.6, which accepts empty string as a valid json.

```
$ php -a
Interactive mode enabled

php > echo phpversion() . "\n"; json_decode('',true); echo json_last_error_msg();
5.6.40-24+ubuntu16.04.1+deb.sury.org+1
No error
php > 

```

```
$ php -a
Interactive mode enabled

php > echo phpversion() . "\n"; json_decode('',true); echo json_last_error_msg();
7.4.3
Syntax error
php > 

```
I'll fix the test and resubmit the PR

tandhika:
I have fixed some broken changes after merging from current master.

src/Provider/AbstractProvider.php
```diff
@@ -98,6 +99,15 @@ abstract class AbstractProvider
...
     protected $optionProvider;
 
+    /**
+     * If a response body cannot be parsed, a value true of this flag will allow
+     * the response parser to throw a ResponseParsingException containing
+     * the response and the body.
+     *
+     * @var bool
+     */
+    protected $mayThrowResponseParsingException = false;
```
ramsey, line 109: I don't like the `$mayThrowResponseParsingException` property. I think it introduces complexity that's not needed. I think there's a better way to preserve BC. See below.

src/Provider/Exception/ResponseParsingException.php
```diff
@@ -0,0 +1,76 @@
...
+
+namespace League\OAuth2\Client\Provider\Exception;
+
+use Exception;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * Exception thrown if the parser cannot parse the provider response.
+ */
+class ResponseParsingException extends Exception
```
ramsey, line 23: What do you think about extending `UnexpectedValueException` instead? This preserves BC, while providing a more specific exception to catch.

src/Provider/Exception/ResponseParsingException.php
```diff
@@ -0,0 +1,76 @@
...
+     * @param ResponseInterface $response The response
+     * @param string $responseBody The response body
+     * @param null $message
+     * @param int $code
+     */
+    public function __construct(
+        ResponseInterface $response,
+        $responseBody,
+        $message = null,
+        $code = 0
```
ramsey, line 45: Add another parameter here for `Throwable $previous = null`.

src/Provider/Exception/ResponseParsingException.php
```diff
@@ -0,0 +1,76 @@
...
+    ) {
+        $this->response = $response;
+        $this->responseBody = $responseBody;
+
+        if (null === $message) {
+            $message = sprintf('Cannot parse response body: "%s"', $responseBody);
+        }
+
+        parent::__construct($message, $code);
```
ramsey, line 54: Pass `$previous` along to the parent here.

src/Provider/AbstractProvider.php
```diff
@@ -686,18 +702,26 @@ protected function parseResponse(ResponseInterface $response)
             return $this->parseJson($content);
         } catch (UnexpectedValueException $e) {
             if (strpos($type, 'json') !== false) {
-                throw $e;
+                throw $this->mayThrowResponseParsingException
+                    ? new ResponseParsingException($response, $content, $e->getMessage(), $e->getCode())
```
ramsey, line 706: Following my other comments about adding `$previous` to the constructor for `ResponseParsingException`, pass `$e` here as the 5th parameter.

src/Provider/AbstractProvider.php
```diff
@@ -686,18 +702,26 @@ protected function parseResponse(ResponseInterface $response)
...
             }
 
-            if ($response->getStatusCode() == 500) {
-                throw new UnexpectedValueException(
-                    'An OAuth server error was encountered that did not contain a JSON body',
-                    0,
-                    $e
-                );
+            // for any other content types
+            if ($this->mayThrowResponseParsingException) {
```
ramsey, line 711: Maybe we always throw the exception, no matter what?

I know this suggestion is somewhat of a BC break, but if you have `ResponseParsingException` extend `UnexpectedValueException`, then anyone catching that exception will start catching the new exceptions and can decide what to do about them.

@shadowhand, @stevenmaguire, thoughts?

ramsey (changes requested):

src/Provider/AbstractProvider.php
tandhika: Agree.

src/Provider/Exception/ResponseParsingException.php
tandhika: Good suggestion

src/Provider/AbstractProvider.php
tandhika: I like your suggestion. I can implement the changes if no objection coming from @shadowhand and @stevenmaguire

src/Provider/Exception/ResponseParsingException.php
tandhika: `Throwable` is introduced in php version 7. Should we keep support for version 5.6, we should use Exception. However, I do see a function used in AbstractProvider which is not supported by 5.6 e.g., random_bytes.

ramsey: Good point on `Throwable`. Use `Exception` for now.

We are continuing to support PHP 5.6 for the 2.x versions. We're able to use `random_bytes()` in 5.6 through a polyfill (`paragonie/random_compat`).

tandhika:
So, I have incorporated reviewer suggestions into the last commit, but I kept the BC until there is a need to change that as well. Thx @ramsey for the suggestions. If I should squash all my commits, please let me know.

PR #833, opened 2020-04-25
Add nextcloud to thirdparty providers

bahuma20:
I've just created a nextcloud plugin by basically copying from the google plugin. So it should be according to best practices.


ramsey:
@bahuma20, thank you for contributing, and I'm sorry it's taken so long to respond. While we don't fully review the code of all third-party providers, we do try to make sure each provider includes tests. Since your package doesn't include tests, I'll close this PR. Feel free to re-submit after you update your package with tests. Cheers!

PR #852, opened 2020-08-06
Allow time to be set for test purposes 

kevinquinnyo:
This just adds a static method for test purposes so that downstream code can more easily assert an exact `AccessToken` object in their tests.

Since the calls to `time()` are non-deterministic, downstream tests are forced to mock, or do looser assertions on the `expires` property.

Usage might be something like this:
```php
// in a test in some codebase that uses this library
public function tearDown()
{
    parent::tearDown();
    AccessToken::resetTimeNow();
}

public function testSomething()
{
    AccessToken::setTimeNow(1000);

    // $something = ... build object under test that uses this library to generate AccessToken
    // $expectedToken = new AccessToken(...)

    $token = $something->getToken();
    $this->assertEquals($token, $expectedToken);
}
```

I did this because I was writing some tests around code that depends on this library, and it seemed useful to me, however I understand if using static methods for test purposes is something you prefer not to do, or it otherwise doesn't fit with the coding preferences here.


kevinquinnyo:
@ramsey I had forgotten about this PR, but I got the notification master had been merged in, so I checked it and noticed that there was a CI failure due to `TestCase::tearDown()` not matching the parent which has a return typehint of `void`.  So I sent up a fix for that real quick.

But fixing this makes the php 5.6 run fail.  Seems like whack-a-mole.  What's the solution here?

ramsey:
I dealt with the same thing getting tests to run on 5.6 through 8.0. The solution is to code around `setUp()` and `tearDown()`. Figure out an alternative to using it. Maybe change the name of the function and then call it at the end of every test method? It sounds silly, but that's probably the best solution for now.

kevinquinnyo:
wack-a-mole complete lol :hammer:

ramsey:
Thanks!

dpi:
@ramsey FWIW there is a strategy that Symfony and Drupal have been using to overcome supporting PHP5 + newer versions of PHPunit:

On test bootstrap, actually modify the test files and remove the return types:

 - https://github.com/symfony/phpunit-bridge/blob/5.x/bin/simple-phpunit.php 
 - https://www.drupal.org/project/drupal/issues/3063887 / https://git.drupalcode.org/project/drupal/blob/5d457e37c5e716551c592c67a70c5cc7bbe2e20a/core/tests/Drupal/TestTools/PhpUnitCompatibility/PhpUnit8/ClassWriter.php#L44

PR #856, opened 2020-09-16
Process options in AbstractProvider->getAccessToken().

### LINKED ISSUE #774: [Question] Client Credentials Grant with scope
Hi folks,

I'm using the package as the "Client Credentials Grant" example from README.md and I'm trying to inform a specific scope called _cbnpagamento_vendas_, appearently without success, example:

```
use League\OAuth2\Client\Provider\GenericProvider;

$options = [
    'clientId' => 'xxx',
    'clientSecret' => 'xxx',
    'urlAuthorize' => 'https://apigw-h.bndes.gov.br/token',
    'urlAccessToken' => 'https://apigw-h.bndes.gov.br/token',
    'urlResourceOwnerDetails' => 'https://apigw-h.bndes.gov.br/cbn-fornecedor',
];

$prov = new GenericProvider($options);
$token = $prov->getAccessToken('client_credentials', ['scope' => ['cbnpagamento_vendas']]);
var_dump($token);
```

And I get:

```
class League\OAuth2\Client\Token\AccessToken#32 (5) {
  protected $accessToken =>
  string(36) "7b27b500-89a2-36b8-8f9c-xxx"
  protected $expires =>
  int(1551893968)
  protected $refreshToken =>
  NULL
  protected $resourceOwnerId =>
  NULL
  protected $values =>
  array(2) {
    'scope' =>
    string(28) "am_application_scope default"       <== HERE!
    'token_type' =>
    string(6) "Bearer"
  }
}
```

If I request the token without specifing the second parameter `['scope' => ['cbnpagamento_vendas']]` I get the same result (same scopes "am_application_scope default"), so it seems to me that parameter doesn't make any difference.

What am I doing wrong?
Thank you!

imclean557:
Fixes #774


imclean557:
The test failures seem unrelated to these changes.

ramsey:
The new test failures are because the mocked method `$grant->prepareRequestParameters($params, $options)` is receiving new values in `$options` that the test does not expect.

A few issues with this PR:

1. There should be a test that shows the problem reported in #774 that this PR fixes
2. I do not believe this change is related to the problem reported in #774 
3. By calling `$this->getAuthorizationParameters()` here, you're merging authorization request parameters onto the access token request, and this could potentially leak credentials into the access token request, which we don't want to happen

For these reasons, I'm closing this PR.

Feel free to open a new PR, if you think there is still a change that needs to be made here. If so, please provide a test that shows the problem you're solving, and please don't call `getAuthorizationParameters()` within `getAccessToken()`, since there's a security risk in doing so.

Thanks!

PR #857, opened 2020-10-02
Update thirdparty.md

Fren1100:
Fix broken link


ramsey:
Please see #670, #660, #638, #679, #671, #662, #691, and #805. See also #773. This works as expected. No PR necessary. :smile:

Issue #862, opened 2020-10-22
PHP 8 support?


angrybrad:
PHP 8 will be officially released in 4 days.

As it currently it, this library won't be installable via Composer if you're running PHP 8 because of https://github.com/thephpleague/oauth2-client/blob/master/composer.json#L9

Was going to attempt to run the tests against PHP 8, but looks like 7.4 isn't even part of the test suite: https://github.com/thephpleague/oauth2-client/blob/master/.travis.yml#L4-L9

Are there plans to add support for PHP 8?

ramsey:
@angrybrad I'm going to take a stab at this to see what's need to support 8. Thanks!

PR #842, opened 2020-05-30
Test against php 7.4

BackEndTea:

PR #864, opened 2020-10-26
Support PHP 8

ramsey:
* Indicate the library supports PHP 7.4 and PHP 8.
* Nothing in `src/` changes
* Update tests to run on PHP 5.6 through PHP 8
  * Replaced Phony with Mockery, since I could not get Phony to work properly on PHP 7.4 and PHP 8
  * Removed the `setUp()` methods because later versions of PHP require the `void` return type on this, while PHP 5.6 would choke on this

This resolves https://github.com/thephpleague/oauth2-client/issues/862


ramsey:
Tests are passing, and there's 100% coverage. All the failures are with Netlify, and I'm not sure what, if anything, needs to be done about that.

ramsey:
FYI, the current failures in the master branch will be resolved when this is merged. I merged #842 to give commit credit to the author, and I didn't realize it would cause the build to fail, since #842 showed it passing.

composer.json
```diff
@@ -6,16 +6,15 @@
         "sort-packages": true
     },
     "require": {
-        "php": "^5.6|^7.0",
+        "php": "^5.6 || ^7.0 || ^8.0",
```
shadowhand, line 9: Probably time to drop support for unsupported versions. As a separate PR, of course.

test/src/Grant/GrantTestCase.php
```diff
@@ -50,36 +46,41 @@ protected function setUp()
...
+        /** @var ClientInterface & MockInterface $client */
+        $client = Mockery::spy(ClientInterface::class)->makePartial();
+        $client
+            ->shouldReceive('send')
+            ->once()
+            ->withArgs(function ($request) {
+                parse_str((string) $request->getBody(), $body);
+                return call_user_func($this->getParamExpectation(), $body);
+            })
+            ->andReturn($response);
```
shadowhand, line 79: Slightly disappointed at the loss of Phony, I find it far more elegant.

shadowhand (approved):

test/src/Grant/GrantTestCase.php
ramsey: Sorry. It appears that it generates code and `eval()`s it to create mocks, and the code it generated kept putting a backslash in front of `?string` type hints, causing syntax errors (i.e. `\?string`). I spent a fair amount of time trying to juggle dependencies to see if I could find the right combination that worked across all PHP versions.

In the end, I decided to change oauth2-client to get it passing on all versions of PHP, so that we could announce PHP 8 support as soon as PHP 8 is out. When Phony supports newer versions of PHP, we can switch back. Switching everything to Mockery didn't take more than an hour, so switching back shouldn't be a big deal.

PR #865, opened 2020-10-27
Add CI workflow for GitHub Actions

ramsey:


ramsey:
Nice! Looks like it won't use a workflow that's not already in the `master` branch, so going ahead and merging this to test it in `master`. 🎉

PR #866, opened 2020-11-08
Allow time to be fetched from a universal clock ⏱

dpi:
What a coincidence I've been putting together an [OAuth project for Drupal](https://drupal.org/project/authman), and in order to complete unit tests I needed a way to mock time. Thanks for the timely implementation of #852 @ramsey @kevinquinnyo 

# Purpose

Allow time to be fetched from a universal clock and eliminate usages of `time()` and `strtotime()` where possible.

Implementing a clock, which is useful for the Drupal [Authman](https://drupal.org/project/authman) project. In Drupal we consider the now-time throughout the lifetime of a request to be the time at the start of the request, not the true current time. So for integration with Drupal I expect this method the proposed clock feature to be used always. For integration with other projects, its also necessary to obey time travel when other projects request it.

# Implementation notes

 - Eliminated remaining calls to `time()` and `strtotime()` where possible. Only 2x calls to time() remain where the true current time is required for getting and testing default time.
 - Added `\League\OAuth2\Client\Token\AccessTokenInterface::getTimeNow` to interface, which was defined in #852
 - Rewrote comments such as "Set the time now. This should only be used for testing purposes.".
 - Switched tests to always simulate time where possible, to eliminate possible random test failures.
 - Given that the work in #852 was designed for tests only, not for public API, I defer to maintainers to whether to remove the static time method stored in `\League\OAuth2\Client\Token\AccessToken::$timeNow` in favor of the clock.

# Clock

Inspiration for the clock comes from a few examples I'm familiar with, such as two popular Clock projects on Packagist, and Drupal.

 - https://github.com/kreait/clock-php/blob/main/src/Clock/SystemClock.php
 - https://github.com/lcobucci/clock/blob/2.1.x/src/SystemClock.php
 - Drupal's https://git.drupalcode.org/project/drupal/-/blob/9.2.x/core/lib/Drupal/Component/Datetime/TimeInterface.php

### DIFF
--- a/src/Provider/Clock.php
+++ b/src/Provider/Clock.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace League\OAuth2\Client\Provider;
+
+/**
+ * Represents an implementation of a Clock.
+ */
+class Clock
+{
+
+  /**
+   * Get the current time.
+   *
+   * @return \DateTimeImmutable
+   */
+    public function now()
+    {
+        return new \DateTimeImmutable();
+    }
+}
--- a/src/Token/AccessToken.php
+++ b/src/Token/AccessToken.php
@@ -15,6 +15,7 @@
 namespace League\OAuth2\Client\Token;
 
 use InvalidArgumentException;
+use League\OAuth2\Client\Provider\ProviderClock;
 use RuntimeException;
 
 /**
@@ -50,12 +51,17 @@ class AccessToken implements AccessTokenInterface, ResourceOwnerAccessTokenInter
     protected $values = [];
 
     /**
-     * @var int
+     * The current time, or NULL to get the true current time via PHP.
+     *
+     * @var int|null
      */
     private static $timeNow;
 
     /**
      * Set the time now. This should only be used for testing purposes.
+
+    /**
+     * Sets the current time.
      *
      * @param int $timeNow the time in seconds since epoch
      * @return void
@@ -66,7 +72,7 @@ class AccessToken implements AccessTokenInterface, ResourceOwnerAccessTokenInter
     }
 
     /**
-     * Reset the time now if it was set for test purposes.
+     * Reset the current time so the true current time via PHP is used.
      *
      * @return void
      */
@@ -76,6 +82,10 @@ class AccessToken implements AccessTokenInterface, ResourceOwnerAccessTokenInter
     }
 
     /**
+
+    /**
+     * Get the current time, whether true or simulated.
+     *
      * @return int
      */
     public function getTimeNow()
@@ -196,7 +206,7 @@ class AccessToken implements AccessTokenInterface, ResourceOwnerAccessTokenInter
             throw new RuntimeException('"expires" is not set on the token');
         }
 
-        return $expires < time();
+        return $expires < $this->getTimeNow();
     }
 
     /**
--- a/src/Token/AccessToken.php
+++ b/src/Token/AccessToken.php
@@ -15,7 +15,7 @@
 namespace League\OAuth2\Client\Token;
 
 use InvalidArgumentException;
-use League\OAuth2\Client\Provider\ProviderClock;
+use League\OAuth2\Client\Provider\Clock;
 use RuntimeException;
 
 /**
@@ -58,7 +58,11 @@ class AccessToken implements AccessTokenInterface, ResourceOwnerAccessTokenInter
     private static $timeNow;
 
     /**
-     * Set the time now. This should only be used for testing purposes.
+     * The clock.
+     *
+     * @var Clock
+     */
+    protected $clock;
 
     /**
      * Sets the current time.
@@ -82,15 +86,25 @@ class AccessToken implements AccessTokenInterface, ResourceOwnerAccessTokenInter
     }
 
     /**
+     * @inheritdoc
+     */
+    public function setClock(Clock $clock)
+    {
+        $this->clock = $clock;
+    }
 
     /**
-     * Get the current time, whether true or simulated.
-     *
-     * @return int
+     * @inheritdoc
      */
     public function getTimeNow()
     {
-        return self::$timeNow ? self::$timeNow : time();
+        if (self::$timeNow) {
+            return self::$timeNow;
+        } elseif (isset($this->clock)) {
+            return $this->clock->now()->getTimestamp();
+        } else {
+            return time();
+        }
     }
 
     /**
@@ -114,6 +128,10 @@ class AccessToken implements AccessTokenInterface, ResourceOwnerAccessTokenInter
 
         if (!empty($options['refresh_token'])) {
             $this->refreshToken = $options['refresh_token'];
+        }
+
+        if (!empty($options['clock'])) {
+            $this->clock = $options['clock'];
         }
 
         // We need to know when the token expires. Show preference to
--- a/src/Token/AccessTokenInterface.php
+++ b/src/Token/AccessTokenInterface.php
@@ -15,6 +15,7 @@
 namespace League\OAuth2\Client\Token;
 
 use JsonSerializable;
+use League\OAuth2\Client\Provider\Clock;
 use RuntimeException;
 
 interface AccessTokenInterface extends JsonSerializable
@@ -69,4 +70,20 @@ interface AccessTokenInterface extends JsonSerializable
      * @return array
      */
     public function jsonSerialize();
+
+    /**
+     * Sets the clock.
+     *
+     * @param Clock $clock a clock.
+     *
+     * @return void
+     */
+    public function setClock(Clock $clock);
+
+    /**
+     * Get the current time, whether real or simulated.
+     *
+     * @return int
+     */
+    public function getTimeNow();
 }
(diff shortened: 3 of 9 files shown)


kevinquinnyo:
I like it

> Given that the work in #852 was designed for tests only, not for public API, I defer to maintainers to whether to remove the static time method stored in \League\OAuth2\Client\Token\AccessToken::$timeNow in favor of the clock.

IMO if this approach is taken, the static method should just be removed in favor of your approach.

1. It was only recently merged, so it's likely no one is utilizing this in their downstream tests (not even me and I authored the PR that added it)
2. It would only affect test code, not production code on the off-chance there were a BC issue.
3. This project already seems to have a lot of BC related juggling as-is, so it seems like a good idea not to introduce yet another one
4. Explicit dependency injection is usually favorable to hidden dependencies (like the static method to fake the time)

I'm not a maintainer here either, so I'll also punt, just my opinion

src/Provider/Clock.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace League\OAuth2\Client\Provider;
+
+/**
+ * Represents an implementation of a Clock.
+ */
+class Clock
```
kevinquinnyo, line 8: Is there a reason not to have something like `ClockInterface` here to require the `now()` method?

src/Token/AccessTokenInterface.php
```diff
@@ -69,4 +70,20 @@ public function __toString();
...
+     * @return void
+     */
+    public function setClock(Clock $clock);
+
+    /**
+     * Get the current time, whether real or simulated.
+     *
+     * @return int
+     */
+    public function getTimeNow();
```
ramsey, line 88: Both of these methods added to the interface represent a BC break.

src/Token/AccessTokenInterface.php
```diff
@@ -69,4 +70,20 @@ public function __toString();
...
     public function jsonSerialize();
+
+    /**
+     * Sets the clock.
+     *
+     * @param Clock $clock a clock.
+     *
+     * @return void
+     */
+    public function setClock(Clock $clock);
```
ramsey, line 81: Typically, we want interfaces to define things we can fetch, not necessarily how they are set. The way something is set could be left open to each implementation, but we'd want to ensure that all implementations have a way to get the clock (i.e., `getClock(): Clock`).

src/Token/AccessTokenInterface.php
ramsey, line 88: If you choose to add `getClock(): Clock`, and the Clock has a `now(): DateTimeInterface` method, then is there any reason to have `getTimeNow()`?

src/Token/AccessToken.php
```diff
@@ -76,11 +86,25 @@ public static function resetTimeNow()
...
      */
     public function getTimeNow()
     {
-        return self::$timeNow ? self::$timeNow : time();
+        if (self::$timeNow) {
+            return self::$timeNow;
+        } elseif (isset($this->clock)) {
+            return $this->clock->now()->getTimestamp();
+        } else {
+            return time();
```
ramsey, line 106: If you implement a `getClock(): Clock` method that must always return a Clock, then there's no need to check for `isset($this->clock)`. Instead, you can do:

``` php
if (self::$timeNow) {
    return self::$timeNow;
}

return $this->getClock()->now()->getTimestamp();
```

src/Provider/Clock.php
```diff
@@ -0,0 +1,20 @@
...
+{
+
+  /**
+   * Get the current time.
+   *
+   * @return \DateTimeImmutable
+   */
+    public function now()
```
ramsey, line 16: Maybe this method should be `getTime()` instead of `now()`, so that Clock could return any DateTime object that an implementer wants to set (fixed, now, etc.)?

src/Provider/Clock.php
dpi: Im all for it but interfaces seem to be used sparingly on this project... I wish providers had an interface.

src/Token/AccessTokenInterface.php
dpi: sorry, I dont understand what you're saying here. If this is about clock interfaces, im all for it.

src/Token/AccessTokenInterface.php
dpi: right, of course

dpi: I dont see how both `getTimeNow` and a theoretical `public getClock()` method would be useful for access tokens. See also my note about whether you'd prefer for the previously implemented time functionality (`getTimeNow`/`setTimeNow`/`resetTimeNow`/`\League\OAuth2\Client\Token\AccessToken::$timeNow`) to be removed.

src/Provider/Clock.php
dpi: I used `now` as thats the same method implemented by the reference clock projects linked in summary. Also a `getTime` is strictly speaking not appropriate as you're getting more than just **time**.

dpi:
I think a lot of this hinges on whether maintainer prefers previous time and clock to co-exist:

 - Remove or keep multiple time methods
 - Add a `public getClock` method? Per above I dont see a valid rationale, yet.

src/Token/AccessTokenInterface.php
ramsey: I'm saying I would prefer to define a `getClock()` method on the `AccessTokenInterface`, rather than a `setClock()` method.

src/Token/AccessTokenInterface.php
ramsey: If you add a `getClock()` method, then I agree that `getTimeNow()` is no longer necessary.

ramsey:
> Add a `public getClock` method? Per above I don't see a valid rationale, yet.

IMO, the rationale for a `getClock()` method is in your purpose statement for this:

> Allow time to be fetched from a universal clock and eliminate usages of `time()` and `strtotime()` where possible

Requiring `setClock()` doesn't enforce anything, since any implementation could treat that as a no-op, especially since properties (i.e., `protected $clock`) aren't part of the contract. If we want to force implementations to respect the contract, then we need to define `getClock(): Clock` that *must* return a `Clock` instance. Then, wherever the `Clock` should be used in the parent classes, we use `getClock()` instead of `$this->clock`. The parent shouldn't care how the `Clock` is set, as long as `getClock()` returns a `Clock`.

src/Provider/Clock.php
shadowhand: I think an interface is appropriate here, because the implementation of `ClockInterface` should be `final class ...`.

src/Provider/Clock.php
shadowhand: `now()` is my preference.

src/Token/AccessToken.php
shadowhand: I think it would be better for `self::$timeNow` to be a `DateTimeImmutable`, rather than an integer.

src/Token/AccessTokenInterface.php
shadowhand: The changes to this interface don't make sense. An access token does not have a "clock", it has an expiry date/time. I think the only reasonable modification here would be `AccessTokenInterface::getExpiresDate()` that returns `DateTimeImmutable`. Given that `getExpires` returns a UNIX timestamp already, it makes no sense for the access token to have a separate clock.

test/src/Provider/ProgrammableClock.php
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * A clock which must be initialised, and may be changed at any time.
+ */
+class ProgrammableClock extends Clock
+{
+
+    /**
+     * @var \DateTimeImmutable|null
+     */
+    protected $time = null;
```
shadowhand, line 16: Why wouldn't this be initialized via `__construct` ?

shadowhand (changes requested):

shadowhand:
The goal is this PR is excellent, and I definitely support the concept of an internal clock. The implementation just needs a little work to be logical.

dpi:
I dont have bandwidth or motivation to work on this right now, but i just want to make-everyone-aware of PSR-20 https://github.com/php-fig/fig-standards/blob/master/proposed/clock.md, which in summary is an interface, with a now(), method, returning a `\DateTimeImmutable` instance. So whatever happens here should aim to be compliant with this, which is basically a standardisation of what other clock projects have arrived at.

Because of this I see it making sense to eventually make PSR be the interface to depend on, so that could mean defining an interface here in oauth2-client, then later making the interface here extend the one from PSR-20.

kevinquinnyo:
> I dont have bandwidth or motivation to work on this right now

Understandable.  I had a little bit of extra time this afternoon so I figured I'd take a stab at it.  I've attempted to apply the code review changes as a PR to your branch @dpi  here: https://github.com/dpi/oauth2-client/pull/1

@ramsey @shadowhand let me know if I've missed anything or if I'm off the mark with those changes.

I think one thing I didn't address is this comment: https://github.com/thephpleague/oauth2-client/pull/866#discussion_r671718866

stof:
This should use the PSR-20 ClockInterface

PR #870, opened 2020-12-31
Adding oauth2-firefly-iii client

stratoss:
Adding a client for [Firefly III](https://www.firefly-iii.org/)


ramsey:
Please include a license file in your repository so that it's clear to users what terms you are granting them. I see "MIT" listed in `composer.json`, but an explicit license file would be better. Thanks!

PR #885, opened 2021-05-12
feat: PSR18

### LINKED ISSUE #685: Considering (again) PHP-HTTP usage
As [suggested on Twitter](https://twitter.com/thephpleague/status/962348828413366272), I'm opening a new issue to ask reconsidering #538 

The pros of using http-plug:

- You don't need to handle the HTTP communication part. Just use the interface, php-http will handle the rest. Less code to write, you can now focus on oauth2 logic only.
- Happy end users (project developers) will not be locked to Guzzle v6 requirement. He can use any version he wants or an another library.
- Want to still use Guzzle 6? Guess what? You can! http://docs.php-http.org/en/latest/clients/guzzle6-adapter.html
- Bench of [plugins](http://docs.php-http.org/en/latest/plugins/) can be used by both maintainers and end users. For example, the end users might want to log the requests thanks to this: http://docs.php-http.org/en/latest/plugins/logger.html

The con of using http-plug: Two more package to add to the `composer require` command line. Not a big deal. Plus, a lot of user are already using php-http so they might not need that. :+1: 

Having the php-http implementation will be very valuable for this library.

Could you please reconsider the PR?

Please see also https://github.com/thephpleague/oauth2-client/pull/538#issuecomment-364661732.

@Nyholm Are you ok to rebase your PR if accepted by the team?

Best regards.

dkarlovi:
Closes #685. See #538.


composer.json
```diff
@@ -7,10 +7,13 @@
...
     "require": {
         "php": "^5.6 || ^7.0 || ^8.0",
-        "guzzlehttp/guzzle": "^6.0 || ^7.0",
+        "psr/http-client-implementation": "^1.0",
+        "psr/http-factory-implementation": "^1.0",
         "paragonie/random_compat": "^1 || ^2 || ^9.99"
     },
     "require-dev": {
+        "guzzlehttp/guzzle": "^7.3",
+        "guzzlehttp/psr7": "^2.0@RC",
```
dkarlovi, line 16: For tests, uses RC because this package doesn't declare providing the PSR17 factories until 2.x.

src/Provider/AbstractProvider.php
```diff
@@ -126,11 +125,7 @@ public function __construct(array $options = [], array $collaborators = [])
         $this->setRequestFactory($collaborators['requestFactory']);
 
         if (empty($collaborators['httpClient'])) {
-            $client_options = $this->getAllowedClientOptions($options);
-
-            $collaborators['httpClient'] = new HttpClient(
-                array_intersect_key($options, array_flip($client_options))
-            );
+            throw new \LogicException('Must specify client');
```
dkarlovi, line 128: The previous workflow no longer works, we cannot create a default client since we don't know what's available.

src/Provider/AbstractProvider.php
```diff
@@ -619,8 +616,8 @@ public function getParsedResponse(RequestInterface $request)
     {
         try {
             $response = $this->getResponse($request);
-        } catch (BadResponseException $e) {
-            $response = $e->getResponse();
+        } catch (ClientExceptionInterface $e) {
+            throw $e;
```
dkarlovi, line 620: There's no equivalent here.

dkarlovi (commented):
@nyholm would love to hear your thoughts.

src/Provider/AbstractProvider.php
Nyholm: One could check if the guzzle client class exists and instantiate if it does. 

That could be a smoother transition but it will be less generic and impartial.

src/Provider/AbstractProvider.php
```diff
@@ -222,7 +217,7 @@ public function setHttpClient(HttpClientInterface $client)
     /**
      * Returns the HTTP client instance.
      *
-     * @return HttpClientInterface
+     * @return ClientInterface
      */
     public function getHttpClient()
```
Nyholm, line 222: Is this method really needed? But it is maybe out of scope for this PR.

src/Provider/AbstractProvider.php
Nyholm: True, But that is a good thing =)

The ClientException will only be thrown when it is impossible to generate a response. Ie network errors or the Request is invalid (missing method) etc.

src/Provider/AbstractProvider.php
```diff
@@ -600,12 +595,14 @@ protected function createRequest($method, $url, $token, array $options)
...
-     * @param  RequestInterface $request
+     * @param RequestInterface $request
+     *
      * @return ResponseInterface
+     * @throws \Psr\Http\Client\ClientExceptionInterface
      */
     public function getResponse(RequestInterface $request)
     {
-        return $this->getHttpClient()->send($request);
+        return $this->getHttpClient()->sendRequest($request);
```
Nyholm, line 605: Maybe add logic here to follow redirects?

Nyholm (commented):
Thank you

src/Provider/AbstractProvider.php
dkarlovi: Yeah, I'd keep it for now, but since this is a BC break, this will have to be 3.x anyway.

src/Provider/AbstractProvider.php
dkarlovi: I agree, was thinking about extracting that logic into a factory, something like

```php
FallbackHttpClient::discover();
```

which could then do some heuristic to find some popular clients. WDYT?

Nyholm: There are 3 ways. 
1) No fallback
2) Just Guzzle. 
3) https://github.com/php-http/discovery

To "find popular clients" is the job of the discovery package.

src/Provider/AbstractProvider.php
dkarlovi: OK, so we can check for either of those and try to construct a client.

src/Provider/AbstractProvider.php
dkarlovi: OK, so we can test for any of those and try to construct the client.

src/Tool/RequestFactory.php
```diff
@@ -23,6 +25,16 @@
...
+    /**
+     * @var RequestFactoryInterface
+     */
+    protected $requestFactory;
+
+    /**
+     * @var StreamFactoryInterface
+     */
+    protected $streamFactory;
+
```
proggeler, line 37: How are these properties injected? I think I'm missing a constructor here.

PR #888, opened 2021-05-25
get a specified value in the access token

hendeavors:
Provide the ability to get a value from the access token by the offset.

Before:

`$value  = $token->getValues()['some_offset'] ?? null;`

After:

`$value = $token->getValue('some_offset');`


src/Token/AccessToken.php
```diff
@@ -207,6 +207,34 @@ public function getValues()
...
+    /**
+     * Determine if the access token has a value by the offset
+     * @param  string|int $offset The offset to check
+     * @return bool
+     */
+    public function hasValue($offset)
+    {
+        if (!(is_string($offset) || is_numeric($offset))) {
+            throw new InvalidArgumentException("Values offset must be a string or int");
+        }
```
shadowhand, line 219: What is the point of this check?

shadowhand (changes requested):

Issue #897, opened 2021-08-02
AbstractProvider - Replace 'AccessToken' dependency with 'AccessTokenInterface' to support cases where AccessToken class is being extended.


cloudcogsio:
A fatal error is thrown for the following case:

1. The concrete class _League\OAuth2\Client\Token\AccessToken_ is extended by a custom provider. Lets call this 'CustomAccessToken'

2. The custom provider overrides the '_createAccessToken_' method to return the '_CustomAccessToken_' class instead of the _League\OAuth2\Client\Token\AccessToken_. (There should be no issue since base functionality is extended and implements '_AccessTokenInterface_'

3. Methods such as '_getResourceOwnerDetailsUrl_' will throw a fatal exception since it's declaration, although abstract, uses the concrete '_League\OAuth2\Client\Token\AccessToken_' instead of the better suited '_AccessTokenInterface_'

**Fix:**
Replace all occurrences of '_League\OAuth2\Client\Token\AccessToken_' with '_League\OAuth2\Client\Token\AccessTokenInterface_' in the abstract methods of **_AbstractProvider.php_**

cloudcogsio:
I've added a new Keycloak OAuth2 client with a branch that implements a use case for extending the base AccessToken.

See https://github.com/cloudcogsio/oauth2-keycloak/blob/master/README.md#custom-access-token-class

ramsey:
Adding this for consideration to our v3 milestone. Thanks!

PR #898, opened 2021-08-03
Changed signature of methods in AbstractProvider.php

cloudcogsio:
- Abstract methods in AbstractProvider.php should use
AccessTokenInterface instead of the concrete AccessToken class provided by League.

- See: https://github.com/thephpleague/oauth2-client/issues/897


shadowhand (commented):
These changes would break all dependent packages due to signature mismatch. This change would have to be done as part of a new major release.

PR #906, opened 2021-09-27
Exclude redirect_uri param from refresh_token requests.

adamroyle:
I ran into this issue with a OAuth2 provider rejecting my refresh_token request as it included the redirect_uri parameter.


ramsey:
Thank you for the PR!

I think this change is best suited as part of a custom provider for the service provider you're using. Best regards!

PR #920, opened 2021-10-28
 PHP 8.1: fix deprecation warnings / http_build_query()

jrfnl:
This fixes an issue with a call to the PHP native `http_build_query()` function, the second parameter of which is the _optional_ `$numeric_prefix` parameter which expects a `string`.

A parameter being optional, however, does not automatically make it nullable.

As of PHP 8.1, passing `null` to a non-nullable PHP native function will generate a deprecation notice.
In this case, this function call yielded a `http_build_query(): Passing null to parameter composer#2 ($numeric_prefix) of type string is deprecated` notice.

Changing the `null` to an empty string fixes this without BC-break.

Fixes nearly all deprecation warnings found when running the tests.

Refs:
* https://www.php.net/manual/en/function.http-build-query.php
* https://wiki.php.net/rfc/deprecate_null_to_scalar_internal_arg

### DIFF
--- a/src/Tool/QueryBuilderTrait.php
+++ b/src/Tool/QueryBuilderTrait.php
@@ -28,6 +28,6 @@ trait QueryBuilderTrait
      */
     protected function buildQueryString(array $params)
     {
-        return http_build_query($params, null, '&', \PHP_QUERY_RFC3986);
+        return http_build_query($params, '', '&', \PHP_QUERY_RFC3986);
     }
 }


ruudk:
@jrfnl Could you maybe rebase this PR to trigger a new CI run? It seems that the tests are stuck. Would be great to have this merged :)

jrfnl:
@ruudk The branch is already at current `master`, so there's nothing to rebase. I think one of the maintainers needs to give permission for the actions to be run or press the re-run button.

ruudk:
Ah, instead of rebasing you can also edit your last commit and force push it, to trigger a new run.

The UI doesn't mention "Waiting for approval to run workflow" so it's not an approval thing.

ruudk:
Weird, now I see the `workflow awaiting approval` but I didn't before. Let's wait then :)

ramsey:
Thanks, @jrfnl!

lyrixx:
Hello, Could you release a new version  with this patch please? It triggers deprecation notice on another projet. Thanks

Slamdunk:
It seems a release is really neaded to use this fix: no `dev` alias is available to match `league/oauth2-client:^2.0` contraints on downstream projects.

PR #928, opened 2021-12-07
add toArray api to access token

hendeavors:
I think this allows some flexibility when working with values provided by the access token. Sometimes it is easier or necessary to pass an array of what the access token provides. I know jsonSerialize provides this already but it doesn't feel right to execute this method for the purposes of working with an array explicitly.

Added an additional interface to avoid a BC for the current release, although I think it would be proper to include the functionality in the next major release.


ramsey:
This looks good to me. Please provide a test.

ramsey (dismissed):
This looks good to me. Please provide a test.

hendeavors:
> This looks good to me. Please provide a test.

Test added.

PR #929, opened 2021-12-17
fix for invalid expires value

hendeavors:
Fixes bug for PHP 7.1+: 

> A non-numeric value encountered

When attempting to add a non-numeric value `$expires` to `getTimeNow` an error is thrown. This should be non-breaking and preserve PHP 5 behavior.


test/src/Token/AccessTokenTest.php
```diff
@@ -189,6 +189,28 @@ public function testInvalidExpiresIn()
...
+    public function testInvalidExpires()
+    {
+        $options = [
+            'access_token' => 'access_token',
+            'expires' => 'TEXT',
+        ];
+
+        $token = $this->getAccessToken($options);
+
+        $this->assertSame($token->getExpires(), $token->getTimeNow());
```
ramsey, line 201: I'm not sure what you're trying to test here, but `'TEXT'` can't be the same as the return value of `getTimeNow()`, which is the current time stamp, so the tests are failing.

src/Token/AccessToken.php
```diff
@@ -121,6 +121,10 @@ public function __construct(array $options = [])
             $expires = $options['expires'];
 
             if (!$this->isExpirationTimestamp($expires)) {
+                if (!is_numeric($expires)) {
+                    $expires = intval($expires);
```
ramsey, line 125: I'm not sure this check is useful. If something is not numeric, it usually won't have an intval worth using.

For example: https://3v4l.org/9r3Wj

ramsey:
Merged to the `master` branch in https://github.com/thephpleague/oauth2-client/commit/8cee80a295a66c6542159596ae0391a557afa7a2

Thank you for contributing! 🎉

PR #946, opened 2022-03-28
Docs: Fix state check example if session variable is not set

Strelox:
The state check in the [Basic Usage](https://oauth2-client.thephpleague.com/usage/) example will only fail if the session variable is set. If an attacker uses a fresh session with the "oauth2state" session variable unset, the check would pass. Even worse if the attacker uses a random state and the check fails, the state variable will be unset in the next line. Then a second request would just pass.


ramsey:
Thank you for contributing! 🎉

PR #959, opened 2022-07-07
Update AbstractProvider.php

jordyk:
`\League\OAuth2\Client\Provider\AbstractProvider::getAuthenticatedRequest` token is nullable


ramsey:
Thank you for contributing! 🎉

Issue #863, opened 2020-10-23
Send additional options to Guzzle


dpi:
Cant pass additional options when creating a request.

For example thephpleague/oauth2-client#803 is trying to use form_params, but `\League\OAuth2\Client\Tool\RequestFactory::getRequestWithOptions` filters them out. Also cannot pass along options in `\League\OAuth2\Client\Provider\AbstractProvider::getResponse

judgej:
The solution I have found, using the Xero provider, is to create my own Guzzle instance outside the provider.

So instead of this:

```php
        $provider = new \Calcinai\OAuth2\Client\Provider\Xero([
            'clientId' => 'my-client-id',
            'clientSecret' => 'my-client-secret',
        ]};
```

I have this:

```php
        $httpClient = new \GuzzleHttp\Client([
            'headers' => [
                'User-Agent' => 'My Xero Application name',
            ],
        ]);

        $provider = new \Calcinai\OAuth2\Client\Provider\Xero([
            'clientId' => 'my-client-id',
            'clientSecret' => 'my-client-secret',
        ], [
            'httpClient' => $httpClient,
        ]};
```

Even though the message is a PSR-7 message that *should* contain everything, Guzzle still slips in some headers on top of that, and the `User-Agent` is one of them. Xero in particular is very keen on the correct `User-Agent` being used, so help with support issues that may arise.

Note that these header parameters are for Guzzle 7. The format is slightly different on Guzzle 6 (wrapped in a `defaults` array) and different again on Guzzle 5.

PR #967, opened 2022-08-11
Add an options argument to the getResponse provider method

matthieuleorat:
To solve the issue https://github.com/thephpleague/oauth2-client/issues/863 I propose the following PR

I wanted to write some tests but feel a bit struggled because i don't know Mockery.


AmirHo3ein13:
I think you also need to add the $option to getParsedResponse() and pass it to getResponse(). For example if you want to pass a json body to the server.

ramsey:
Closing this in favor of focusing on https://github.com/thephpleague/oauth2-client/pull/1032 for the next major version of this library.

Thank you for contributing! 🎉

Issue #976, opened 2022-10-20
ID Token for access token possible?


t3touch:
How can I add an ID Token to the access tokens?

t3touch:
I think I got it 😃 
Or is another way recommended?

```
// Try to respond to the request
$result = $server->respondToAccessTokenRequest($request, $response);
$result->getBody()->write(json_encode(['id_token' => 'test'])); // this is the line that adds the id_token to the response
```

PR #899, opened 2021-08-03
OpenID Connect Discovery Support

cloudcogsio:
Initial Commit
Tests and Documentation to follow.

PR #981, opened 2022-11-12
Abstract out tokens from being tied tightly to access

bradjones1:
Some OAuth2 clients also support OpenID Connect (OIDC.) Also, some native app integrations (e.g., Google) directly return OIDC ID tokens to the native app, which then can be used by a backend for SSO. You get the idea.

There are some other issues/PRs open for OIDC functionality, e.g. https://github.com/thephpleague/oauth2-client/pull/899, and those are good starts for sure. I think what is missing from at least that PR, and more generally, is the ability to represent tokens in a more abstract way. E.g., an OIDC ID token represents a resource owner, but does not contain an access token. This can also lead to some confusion like in https://github.com/thephpleague/oauth2-client/issues/976.

See also https://github.com/thephpleague/oauth2-google/pull/119 for the Google provider, which is an example of how these libraries can be augmented with ID token support to better serve as a backend for all types of SSO implementations, not just browser-based OAuth2 flow.


bradjones1:
I realize the code style is a bit messed up here but am jamming on a project and figure if maintainers are into this, it's easy enough to clean up. Would perhaps be nice to ship some composer commands to do CS linting and fixing locally.

PR #983, opened 2022-12-05
- thirdparty: mkaverin/oauth2-donationalerts

mish-ka-mishka:
Added DonationAlerts provider client to the provider client list


ramsey:
Thank you for contributing! 🎉

PR #993, opened 2023-02-24
hasExpired() should use self::getTimeNow()

prikkprikkprikk:
While trying to test some code that caches an AccessToken and refreshes it if it has expired, I discovered that AccessToken->hasExpired() uses time() instead of self::getTimeNow(), and thus ignores self::$timeNow when testing.

This pull request fixes that.


src/Token/AccessToken.php
```diff
@@ -196,7 +196,7 @@ public function hasExpired()
             throw new RuntimeException('"expires" is not set on the token');
         }
 
-        return $expires < time();
+        return $expires < self::getTimeNow();
```
ramsey, line 199: Should this be `$this->getTimeNow()`, since `getTimeNow()` is an instance method and not a static method?

ramsey:
Merged to the `master` branch in https://github.com/thephpleague/oauth2-client/commit/3c8c2b4b13f4a3ebd770c259cb606ba93871030a

Thank you for contributing! 🎉

Issue #658, opened 2017-09-20
Refresh token lost on renewal


ghost:
Hello,

The refresh token is lost when used to renew the access token and no refresh token is sent with the access token.

Library needs to include the previous (Current) refresh token with the new access token if no new refresh token is provided.

ramsey:
Can you provide more information?

If you're encountering a problem with the refresh token or access token, can you show some snippets of code and explain what you're expecting to happen and what is actually happening?

If you're proposing a feature, can you show some example code to illustrate what you're proposing?

Thanks!

ChrisThompsonTLDR:
I'm having the same issue.  The refresh token is lost after the first refresh.

```
$accessToken = cache('accessToken');

if ($accessToken->hasExpired()) {
    $reddit = new Reddit([
        'clientId'      => 'myid',
        'clientSecret'  => 'mysecret',
        'redirectUri'   => 'http://mydomain.com/callback',
        'userAgent'     => 'laravel:athing:0.1, (by /u/me)',
    ]);

    try {
        $newToken = $reddit->getAccessToken('refresh_token', [
            'refresh_token' => $accessToken->getRefreshToken(),
        ]);
    } catch (Exception $e) {
        dd($e->getMessage());
    }

    Cache::forever('accessTokenOld', $accessToken);

    Cache::forever('accessToken', $newToken);
}
```

After the first refresh, the accessToken looks like
```
League\OAuth2\Client\Token\AccessToken Object
(
    [accessToken:protected] => jkfldajsfd90r93j2fjDKL
    [expires:protected] => 1520698853
    [refreshToken:protected] => 
    [resourceOwnerId:protected] => 
    [values:protected] => Array
        (
            [token_type] => bearer
            [device_id] => None
            [scope] => modposts read submit
        )

)
```

The initial/first accessToken looks like this (I moved it to another cache key)

```
League\OAuth2\Client\Token\AccessToken Object
(
    [accessToken:protected] => jdf()9fJDFkvljf390290sdf
    [expires:protected] => 1520694920
    [refreshToken:protected] => 5639954-IF(d0sjfi032jilsajfd90
    [resourceOwnerId:protected] => 
    [values:protected] => Array
        (
            [token_type] => bearer
            [scope] => modposts read submit
        )

)
```

ChrisThompsonTLDR:
[...]

PR #994, opened 2023-02-26
Support setting refreshToken

grimmdude:
In the cases where the refresh token isn't included in the response when using it to refresh an access token, it's often necessary to persist the original refresh token for later use.  When this happens it's convenient to store the refresh token in the `AccessToken` object as it was originally.

However, because `AccessToken::$refreshToken` is a protected property it's not possible to assign this value directly.  Implementing `AccessToken::setRefreshToken()` allows the refresh token to be set and stored along with the `AccessToken` object.

The workaround I've found is to serialize the `AccessToken` object to JSON, add the refresh token, then reinstantiate a new `AccessToken` object.  It would be great if this wasn't necessary.

Building the functionality directly into the library to automatically carry over the refresh token would be great (as suggested in #658), but I'm not sure if that behavior would apply to all cases.  It might cause confusion if it was intentionally omitted from the response because it's no longer valid.

Related to https://github.com/thephpleague/oauth2-client/issues/658

### DIFF
--- a/src/Token/AccessToken.php
+++ b/src/Token/AccessToken.php
@@ -172,6 +172,14 @@ class AccessToken implements AccessTokenInterface, ResourceOwnerAccessTokenInter
     /**
      * @inheritdoc
      */
+    public function setRefreshToken($refreshToken)
+    {
+        $this->refreshToken = $refreshToken;
+    }
+
+    /**
+     * @inheritdoc
+     */
     public function getExpires()
     {
         return $this->expires;
--- a/src/Token/AccessToken.php
+++ b/src/Token/AccessToken.php
@@ -22,7 +22,7 @@ use RuntimeException;
  *
  * @link http://tools.ietf.org/html/rfc6749#section-1.4 Access Token (RFC 6749, §1.4)
  */
-class AccessToken implements AccessTokenInterface, ResourceOwnerAccessTokenInterface
+class AccessToken implements AccessTokenInterface, ResourceOwnerAccessTokenInterface, SettableRefreshTokenInterface
 {
     /**
      * @var string
--- a/src/Token/ResourceOwnerAccessTokenInterface.php
+++ b/src/Token/ResourceOwnerAccessTokenInterface.php
@@ -14,7 +14,7 @@
 
 namespace League\OAuth2\Client\Token;
 
-interface ResourceOwnerAccessTokenInterface
+interface ResourceOwnerAccessTokenInterface extends AccessTokenInterface
 {
     /**
      * Returns the resource owner identifier, if defined.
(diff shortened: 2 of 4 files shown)


src/Token/AccessToken.php
```diff
@@ -169,6 +169,14 @@ public function getRefreshToken()
         return $this->refreshToken;
     }
 
+    /**
+     * @inheritdoc
+     */
+    public function setRefreshToken($refreshToken)
```
ramsey, line 175: Since this isn't part of any interface, if the calling code has type-hinted for the `AccessTokenInterface` or the `ResourceOwnerAccessTokenInterface`, then there's no guarantee that the access token object will have this method.

Rather than changing the existing interfaces, which would break BC, I propose introducing a new interface. Something like `SettableRefreshTokenInterface`. It would have a simple API:

```php
namespace League\OAuth2\Client\Token;

interface SettableRefreshTokenInterface
{
    /**
     * Sets or replaces the refresh token with the provided refresh token
     *
     * @param string $refreshToken
     *
     * @return void
     */
    public function setRefreshToken($refreshToken);
}
```

Then, on `AccessToken`, we would implement this interface.

If anyone wants to hint for both `AccessTokenInterface` and `SettableRefreshTokenInterface`, they may use intersection types (i.e., `AccessTokenInterface & SettableRefreshTokenInterface`) either in native code (PHP 8.1+) or in docblock annotations (for PHPStan, Psalm, etc.).

Thoughts?

grimmdude: Thanks for your comment @ramsey.  Good point, I hadn't considered the effects of this method not being defined in any interface.

Creating a new `SettableRefreshTokenInterface` interface like you suggest sounds like the best bet to avoid breaking changes.  Keeping it in `AccessTokenInterface` _would_ be nice since `getRefreshToken()` is already defined there, but maybe something that can be consolidated in the next major release?

I've just committed that new `SettableRefreshTokenInterface` ✅

ramsey: Thanks!

ramsey (approved):

Seldaek:
Heya @ramsey @shadowhand could I bother you for a release with this PR? A few commits are missing from 2.7.0 https://github.com/thephpleague/oauth2-client/compare/2.7.0...master - thanks 🙏🏻

Seldaek:
@ramsey @shadowhand sorry to nag again but it's been a year and still no release :)

Seldaek:
And if someone does a release please merge https://github.com/thephpleague/oauth2-client/pull/1039 first so we get PHP 8.4 compatibility sorted as well :)

Issue #935, opened 2022-01-24
how do i pass $_POST parameters with "getAuthenticatedRequest()"?


ivan006:
Seems I can only pass php://input payload

ivan006:
oh i see i must just do this

$body = http_build_query($articulation);
$headers["content-type"] = 'application/x-www-form-urlencoded';

fverry:
Also see https://github.com/thephpleague/oauth2-client/pull/967 / https://github.com/thephpleague/oauth2-client/pull/995

Issue #985, opened 2023-01-24
Create multipart/form-data requests?


Lindstromer:
I'm having issues creating a request with an attached file. Tried a lot of different alternatives but always end up with issues.

Seems like Guzzle in itself handles this, but the oauth2-client is limiting the getAuthenticatedRequest method to using only " Any of "headers", "body", and "protocolVersion".".

Example where it says to not specify form-data:
https://stackoverflow.com/questions/47550801/multiple-files-uploaded-via-guzzle-multipart-form-data-request-are-not-recognize

Another discussion regarding this, I tried following it but cannot get it to work. Guzzle complains about it being an invalid resource type "array" while doing so:
https://stackoverflow.com/questions/58736361/replicate-curl-multipart-form-data-request-with-guzzle-league-oauth2-client?rq=1

Error is thrown in the streamFor method found in Utils.php located in guzzlehttp vendor/psr7/src/Utils.php.

Anyone tried this and got it to work? Any help would be very appreciated.

fverry:
You can make it work with this following extra code :
```php
if (isset($options['multipart']))
{
	$multipartStream = new GuzzleHttp\Psr7\MultipartStream($options['multipart']);
	$boundary = $multipartStream->getBoundary();
	$options['headers']['content-type'] = "multipart/form-data; boundary={$boundary}";
	$options['body'] = $multipartStream;
	unset($options['multipart']);
}
```

It is inspired from [`GuzzleHttp\Client::applyOptions()`](https://github.com/guzzle/guzzle/blob/7.5.0/src/Client.php#L340) which you should read.

As stated by this [stackoverflow answer](https://stackoverflow.com/a/59040489/598259), the only keys allowed in [`League\OAuth2\Client\Provider\AbstractProvider::getAuthenticatedRequest()`'s `$options`](https://github.com/thephpleague/oauth2-client/blob/2.6.1/src/Provider/AbstractProvider.php#L571) are `"headers"`, `"body"`, and `"protocolVersion"`.

Thus, in the following code will ignore the `"multipart"` key in the `$options` array.
```php
$multipart = [];
$multipart[] = [
	'name'     => "documents[{$key}]",
	'contents' => GuzzleHttp\Psr7\Utils::tryFopen($file['tmp_name'], 'r'),
	'filename' => $file['name'],
];
$options['multipart'] = $multipart;
$request = $provider->getAuthenticatedRequest($method, $url, $token, $options);
```

fverry:
[...]

PR #995, opened 2023-03-25
Fix: sending JSON or multipart requests

fverry:
Sending url-encoded params, json or multipart requests with Guzzle is easy, because shortcuts are implemented within [`GuzzleHttp\Client::applyOptions()`](https://github.com/guzzle/guzzle/blob/7.5.0/src/Client.php#L340).

Currently, those shortcuts are unreachable through the League OAuth2 Client, and thus all its providers.

This fix relies on `AbstractProvider::getResponse()`, and does not affect `AbstractProvider::getAuthenticatedRequest()`.

[`GuzzleHttp\ClientInterface::send()`](https://github.com/guzzle/guzzle/blob/7.5.0/src/ClientInterface.php#L30) method supports a second optional `$options` parameter [since v6.0](https://github.com/guzzle/guzzle/commit/1a9ad6b55368283aaf73672a519e5177530991fb?diff=split#diff-7fe9d7bab1528c58fb39365621c6f02430dd01828c26fe8b00f0408de72c71b8).

Upgrading `AbstractProvider::getResponse()` to support this second optional parameter allows the developper to use a `"form_params"`, `"json"` or `"multipart"` option key :
```php
$options = ['multipart' => $multipart];
$request = $oauth2client->getAuthenticatedRequest($method, $url, $token);
$response = $oauth2client->getResponse($request, $options);
```
----
### See also
Fixes: #935 #985 
Duplicates: #967

----
### For immediate use

Run the following on composer CLI:
```sh
composer config repositories.oauth2-client vcs https://github.com/fverry/oauth2-client
composer require "league/oauth2-client:dev-httpclientinterface-send-options as 2.x-dev"
```

----
### Workaround

Use:
```php
$request = $oauth2client->getAuthenticatedRequest($method, $url, $token);
$oauth2client->getHttpClient()->send($request, $options);
```
Instead of:
```php
$request = $oauth2client->getAuthenticatedRequest($method, $url, $token, $options);
$oauth2client->getResponse($request);
```


ramsey:
@fverry, thank you for submitting this PR. Please take a look at the failing tests and fix those. Then, we'll accept this.

isleshocky77:
I'd note that this is an issue for `query` parameters as well as using `json`.  

@fverry I'm wondering why you chose this as the fix though. Would it not be better to fix [RequestFactory::getRequestWithOptions](https://github.com/thephpleague/oauth2-client/blob/master/src/Tool/RequestFactory.php#L75) to use [GuzzleHttp\Client::applyOptions()](https://github.com/guzzle/guzzle/blob/7.5.0/src/Client.php#L340) instead of cherry-picking only `headers`, `body`, and `version`?

gavin310:
Wow thank god I found this thread because I was having a hell of a time getting POST requests to work. `$oauth2client->getHttpClient()->send($request, $options);` saved the day.

ramsey:
Closing this in favor of focusing on https://github.com/thephpleague/oauth2-client/pull/1032 for the next major version of this library.

Thank you for contributing! 🎉

PR #997, opened 2023-04-03
Improve type of $options parameter in Provider\AbstractProvider::getAccessToken()

yguedidi:
Avoid contravariance issue on high levels of static analysis tools for children classes


ramsey:
Thank you!

PR #1002, opened 2023-05-03
Add Disqus provider

antalaron:
I've just created a [Disqus](https://disqus.com) provider.

Thanks for this library.


ramsey:
Thank you for contributing! 🎉

PR #1004, opened 2023-05-28
Replace EVE Online provider.

tkhamez:
Note that the provider I replaced no longer exists on GitHub.


tkhamez:
No response after more than a year, and the other provider was resurrected.

PR #1009, opened 2023-08-28
Update AbstractProvider.php - Fixing return type as Psalm complains a…

michelgokan:
…bout it

This function returns an instance of type AccessToken instead of AccessTokenInterface. The latest version of `psalm` is complaining about this issue.


src/Provider/AbstractProvider.php
```diff
@@ -837,7 +837,7 @@ protected function prepareAccessTokenResponse(array $result)
      *
      * @param  array $response
      * @param  AbstractGrant $grant
-     * @return AccessTokenInterface
+     * @return AccessToken
```
ramsey, line 840: Since `AccessToken` implements `AccessTokenInterface`, I don't understand why Psalm would complain about this.

Is this still a problem in the latest version of Psalm?

PR #1016, opened 2023-11-15
Add hint for session start to usage

pkess:
When trying to use the code from the documentation i got some errors that the session was not started. Maybe the added line here could help others to integrate this.


ramsey:
Thank you for contributing! 🎉

PR #1026, opened 2024-02-15
Fix tests by returning array from PSR7's ResponseInterface::getHeader()

KorvinSzanto:
PSR-7's `ResponseInterface` [returns an array](https://github.com/php-fig/http-message/blob/master/src/MessageInterface.php#L90-L94) from `getHeader` and not a string. This PR updates tests to return an array as well.


ramsey:
Thank you for contributing! 🎉

PR #1036, opened 2024-09-09
Dependabot: submit PRs for GH action runners

jrfnl:
I noticed a number of action runners used are out of date. Most updates are related to the Node version the action runners use under the hood updating to Node 20. Not having those updates will become problematic soonish when GHA drops support for Node 16.

So, instead of manually updating the workflows, I've added the GHA environment to the Dependabot configuration. This should take care of it via Dependabot opening PRs to do the updates.


jrfnl:
Note: the test failure is unrelated to this PR.

ramsey:
Thank you for contributing! 🎉

PR #1033, opened 2024-07-15
Add php 8.3 to test matrix

flavioheleno:
This PR adds PHP 8.3 to the testing matrix.

PR #1039, opened 2024-10-31
Make compatible with PHP 8.4

ruudk:
> League\OAuth2\Client\Provider\AbstractProvider::authorize(): Implicitly marking parameter $redirectHandler as nullable is deprecated, the explicit nullable type must be used instead

When https://github.com/thephpleague/oauth2-client/pull/1033 is merged, I will propose a PR to add PHP 8.4 to the matrix as well.


ruudk:
@ramsey Could you please approve the workflow?

carloscz:
hi, running our internal PHP 8.4 workflows and this abstract library emits a lot of deprecations for us so thank You very much for this PR :) is there any timeframe for release?

ruudk:
@shadowhand Would you be able to help out?

stephanvierkant:
https://github.com/hwi/HWIOAuthBundle looks unmaintained so I'm looking for an alternative, but this repo looks unmaintained as well :(

ruudk:
I decided to remove this dependency and do the OAuth dance with Google myself. Wasn't that hard.

carloscz:
yeah, we probably will switch to JWT parsing and signature checking only as well if this doesn't get pulled through

stephanvierkant:
About this PR: let's drop old PHP versions. Even PHP 8.0 is end of life.

ramsey:
I approved the workflow, but now there are a bunch of failures that appear unrelated to your change. I'll need to investigate and fix these.

PR #1050, opened 2024-12-12
add missing exception annotations

asteragos:
This PR adds missing _@throws_ annotations to method doc blocks in the AbstractProvider class, ensuring all thrown exceptions are properly documented.

**Related Issue**
Fixes [Issue #880](https://github.com/thephpleague/oauth2-client/issues/880) by improving PHPDoc accuracy, enhancing developer experience, and supporting static analysis tools.

**Changes**
Added @throws annotations for all undocumented exceptions in method doc blocks.
Verified annotations match the exceptions thrown.

PR #1055, opened 2025-01-17
add missing guzzle-exception annotations

asteragos:
This PR enhances exception documentation by adding missing throws-annotations for guzzle exceptions like ConnectException, which can be thrown across the package. It builds upon the changes made in #1050.

**Changes**
Added throws-annotations for guzzle exceptions in method doc blocks.


ramsey:
Thank you for contributing! 🎉

PR #1064, opened 2025-04-20
added DeviceCode Grant (see rfc 8628)

slemke76:
added device_code grant - see rfc8628


slemke76:
tests missing, sorry.

PR #1089, opened 2026-02-18
Replace VKontakte Provider

yasinovsky:
In May 2026, the OAuth 2.0 protocol, implemented in the j4k/oauth2-vkontakte library, will cease to function. The new VKontakte protocol will be a OAuth 2.1 with PKCE Challenge. The j4k library has been unsupported for over 10 years and is no longer accepting requests. Due to my work at the company, I work with the VKontakte API and am ready to support the new library on an ongoing basis.


Sephster:
Seems like a sensible change. Thanks for agreeing to take this on @yasinovsky

PR #1091, opened 2026-02-22
Replace 3rd-party Microsoft provider with maintained version

tunterreitmeier:
It seems the current recommended provider is abandoned for almost 10 years now.
That's why I would love to contribute a maintained version. It would also add support for OpenID Connect tokens.


Sephster:
Thanks @tunterreitmeier

PR #1112, opened 2026-07-21
Add support for Guzzle 8

GrahamCampbell:
Guzzle 8.0.0 was released on July 20th. Lots of people want to use it, and thus popular packages in the ecosystem like this one need to be updated to support it, so here I am, adding support. :)


Sephster:
Thanks Graham - will get a release tagged soon

